PR #7, opened 2024-07-11
Rust application

5herlocked:
*Issue #, if available:*
Rust Microservice source code

*Description of changes:*
Acme Store is a sample application built using Next.JS and React.JS along with an API Gateway + Rust Backend.

Tech Stack: Rust, API Gateway, DynamoDB

### DIFF
--- a/applications/rust/buildspec.yaml
+++ b/applications/rust/buildspec.yaml
@@ -13,3 +13,8 @@ phases:
       - cd integration
       - docker build -f Dockerfile -t $ECR_REGISTRY/$ECR_REPOSITORY:$CODEBUILD_RESOLVED_SOURCE_VERSION .
       - docker push $ECR_REGISTRY/$ECR_REPOSITORY:$CODEBUILD_RESOLVED_SOURCE_VERSION
+
+  integration:
+    commands:
+      - cd ./application/rust/
+      - cargo test # @vshardul write some tests brother. Unit tests are usually quite nice
--- a/applications/rust/deployment/infra/rust-app-pod-identity-setup.yaml
+++ b/applications/rust/deployment/infra/rust-app-pod-identity-setup.yaml
@@ -1,0 +1,86 @@
+apiVersion: iam.aws.upbound.io/v1beta1
+kind: Role
+metadata:
+  labels:
+    addToInstanceProfile: "true"
+  name: crossplane-iam-role
+spec:
+  forProvider:
+    assumeRolePolicy: |
+      {
+        "Version": "2012-10-17",
+        "Statement": [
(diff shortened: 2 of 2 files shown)


applications/rust/buildspec.yaml
```diff
@@ -0,0 +1,20 @@
+version: 0.2
+
+phases:
```
shapirov103, line 3: You need to integrate with Gitea, otherwise no way to build it in the ws.

applications/next-js/amplify.yml
```diff
@@ -0,0 +1,18 @@
+version: 1
```
shapirov103, line 1: How do we deploy this? Just CDN distribution for SPA?

applications/rust/buildspec.yaml
5herlocked: Ack, I was under the assumption that we were going to use code build. Will work with Zach/Renuka to convert this to Gitea

applications/next-js/amplify.yml
5herlocked: Let me get back to you on this exactly.

5herlocked: https://nextjs.org/docs/pages/building-your-application/deploying#docker-image -- will modify this to use docker and k8s

shapirov103: Let's add dockerfile

applications/rust/buildspec.yaml
shapirov103: we are using codebuild, but how do you think it will get the source? I will have to use nosource option and use git commands to check it out.

applications/rust/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+- `GET /menu/<menu_id>`: Retrieves the menu structure (categories and collections).
+- `GET /pages`: Retrieves a list of all pages -- **UNUSED**.
+- `GET /page/<_page_handle>`: Retrieves details of a specific page -- **UNUSED**.
+- `GET /category/<category_handle>`: Retrieves details of the category
+- `GET /category/<category_handle>/products`: Retrieves the products in that category
+- `GET /categories`: Retrieves a list of all the categories
+- `GET /collection/<collection_handle>`: Reserved for special collections for events/front_page
+
+### To Do for Workshop
```
shapirov103, line 38: Have you tried doing that? I doubt that customers will have working knowledge of rust, we need to really babysit them. I like the use case when q dev allows them to enter rust arena without much knowledge, but don't want to go down the rabbithole during the workship with people trying to use a pointer or something,

PR #13, opened 2024-08-13
add terraform for eks cluster import and irsa

zjaco13:
*Issue #, if available:*

*Description of changes:*
Created terraform for importing an EKS cluster, creating a namespace, then adding a service account to that namespace with permissions through irsa.

### DIFF
--- a/platform/backstage/templates/import-cluster-with-irsa/templates-cluster-irsa/manifests/gitrepository.yaml
+++ b/platform/backstage/templates/import-cluster-with-irsa/templates-cluster-irsa/manifests/gitrepository.yaml
@@ -6,6 +6,8 @@ metadata:
   annotations:
     argocd.argoproj.io/sync-wave: "1"
 spec:
+  secretRef:
+    name: appmod-blueprints-secret
   interval: 30s
   url: https://github.com/aws-samples/appmod-blueprints.git
   ref:


platform/infra/terraform/irsa-example/main.tf
```diff
@@ -0,0 +1,136 @@
...
+resource "kubernetes_namespace" "namespace" {
+  metadata {
+    name = var.namespace
+  }
+}
+
+data "tls_certificate" "eks" {
+
+  url = data.aws_eks_cluster.eks_cluster.identity[0].oidc[0].issuer
+
```
elamaran11, line 47: Does this work even if the user doesnt pass a tls_cert?

platform/infra/terraform/irsa-example/main.tf
```diff
@@ -0,0 +1,136 @@
...
+  assume_role_policy = <<POLICY
+
+{
+
+  "Version": "2012-10-17",
+
+  "Statement": [
+
+    {
```
elamaran11: Nit PIcky. could you align this better?

platform/infra/terraform/irsa-example/main.tf
zjaco13: The user doesn't need to specify anything for this, it pulls directly from the eks cluster, but this resolved issues I was having creating an oidc provider with the thumbprint list.

elamaran11: Make sense.

platform/infra/terraform/irsa-example/main.tf
zjaco13: changed the spacing so its a little more clear, from looking at other iam policies it looks like each statement is supposed to be tabbed in once from the statement field

elamaran11: LGTM

platform/infra/terraform/irsa-example/main.tf
```diff
@@ -0,0 +1,119 @@
+terraform {
```
shapirov103, line 1: So import cluster and creating a service account in it requires access to the cluster using the role that is registered as system:masters (e.g. aws-auth config map). How do you know which role was used to create the cluster so that you can import it correctly, is it a convention of some sort? Are you using access entries for authentication?

I assume that you just happened to use the same principal for IRSA that was used to create the cluster in the first place, that may be a constraint that we will have to adopt. To support import of clusters provisioned in other accounts by other principals we will need to think through this.

zjaco13: Yeah, creating and importing were done directly from my account, so I'm not sure about other principles/roles.

PR #17, opened 2024-08-28
Changes to fetch and display orders and OTEL instrumentation

renukakrishnan:
*Description of changes:*
1. Added server side and UI code to fetch and display the orders
2. Added AWS OTEL instrumentation for metrics and traces


applications/dotnet/src/WebUI/otel-collector-config.yml
```diff
@@ -0,0 +1,38 @@
...
+  otlp:
+    protocols:
+      grpc:
+        endpoint: 0.0.0.0:4317
+      http: null
+processors:
+  batch: null
+exporters:
+  prometheusremotewrite:
+    endpoint: <YOUR-APS-WORKSPACE-ENDPOINT>/api/v1/remote_write
```
shapirov103, line 11: Is this correct config? Who is going to populate the endpoint?

shapirov103: @renukakrishnan ^^

shapirov103: @elamaran11 I assume we will have an approach that enables us to scape metrics from the app without spreading the knowledge of the remote write endpoint. It is not going to scale and requires changes between dev and prod. 

@renukakrishnan please check if there is an approach to emit OTEL metrics from the app in a way that can be scraped by the OTEL collector.

elamaran11: @shapirov103 Yes Dev team should not be doing this, we have already wired collector config from Platform side on both Dev and Prod clusters to send metrics to AMP. Yes Dev team doing at app leve is not scalable approach.

renukakrishnan: @shapirov103 I have added the internal DNS name of the OTEL collector. Also attaching a file that shows the metrics that will get emitted.[
[prometheus_exporter_data.txt](https://github.com/user-attachments/files/17282756/prometheus_exporter_data.txt)
](url)

applications/dotnet/src/Persistence/DependencyInjection.cs
```diff
@@ -9,8 +9,23 @@ public static class DependencyInjection
     {
         public static IServiceCollection AddPersistence(this IServiceCollection services, IConfiguration configuration)
         {
+            var configurationbuilder = new ConfigurationBuilder()
+                .AddJsonFile("appsettings.json")
+                .AddJsonFile($"appsettings.Local.json", optional: true)
+                .AddKeyPerFile("/opt/secret-volume", optional: true, reloadOnChange: true)
```
shapirov103, line 18: Nice!

shapirov103 (approved):
LGTM

PR #27, opened 2024-09-24
Update backstage templates and add vela templates

zjaco13:
*Issue #, if available:*

*Description of changes:*
Adds create dev and prod cluster template and removes cnoe references

### DIFF
--- a/platform/backstage/README.md
+++ b/platform/backstage/README.md
@@ -2,6 +2,7 @@
 
 This repo has following backstage templates which can be plugged in to your Internal Developer Portal via backstage readily:
 
+- `create-dev-and-prod-env` backstage template enables you to deploy 2 EKS clusters for Dev and Prod application environments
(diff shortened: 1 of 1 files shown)


platform/traits/service-account.cue
```diff
@@ -0,0 +1,93 @@
...
+	description: "Specify a ServiceAccount for your workload to get access to components and other resources with IAM Roles"
+	attributes: {
+		podDisruptive: false
+		appliesToWorkloads: ["deployments.apps", "statefulsets.apps", "daemonsets.apps", "jobs.batch"]
+	}
+}
+template: {
+	parameter: {
+    // +usage=Specify the components with policies to add to the service account
+    componentNamesForAccess?: [...string]
```
shapirov103, line 14: is this something that we are moving forward with? I believe the result of your experiment showed that you can't access all components in the app based from the trait.

zjaco13: This isn't used to access the cluster components, it is just used to get the name of the policy that is attached to that component with the component-policy trait, which is based on that components name

PR #39, opened 2024-10-07
Fixing Guest from Teams for backstage templates

elamaran11:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/platform/backstage/templates/catalog-info.yaml
+++ b/platform/backstage/templates/catalog-info.yaml
@@ -7,8 +7,8 @@ spec:
   targets:
     - ./rds-cluster/template.yaml
     - ./s3-bucket/template.yaml
-    - ./microservices/template.yaml
-    - ./microservices-with-repo/template.yaml
+    - ./app-deploy/template.yaml
+    - ./app-deploy-without-repo/template.yaml
     - ./jupyterhub-on-eks/template-jupyterhub-on-eks.yaml
     - ./spark-on-eks/template-data-on-eks.yaml
     - ./eks-observability-accelerator/template-infra-monitoring.yaml


platform/backstage/templates/microservices-with-repo/template.yaml
```diff
@@ -35,6 +35,7 @@ spec:
       action: cnoe:create-argocd-app
       input:
         appName: ${{parameters.microserviceName}}
+        appNamespace: default
```
shapirov103: why default? should be team namespace. 
also i don't see environment specification. Somewhere there should be logic to create the argo app with the destination of the dev server.

elamaran11: This is Argo App Namespace and not the microservices namespace, this could be default and thats what we use across all templates but if you insist it to be a namespace of user choice. How are we going to make sure the namespace exist? 
And i think destination is a new requirement which we have not tried with current backstage implementation. Im not sure if that feature is supported with cnoe:create-argocd-app method. We should research. I would say merge this and move forward with your POC. We can work on rest in remaining.

shapirov103: our requirements are the same: we need apps deployed in dev and prod. If I cannot accomplish this with this template, I will have to use something else.

shapirov103: I suppose just adding env parameter and setting the destination of the app to that parameter should accomplish that.

elamaran11: It is done already. Pls check.

platform/backstage/templates/microservices-with-repo/template.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: scaffolder.backstage.io/v1beta3
 kind: Template
 metadata:
   name: microservices-without-git-repo
-  title: MicroServices with Git Repo
+  title: MicroServices without Git Repo
```
shapirov103: This is causing cognitive dissonance: the folder name is "microservices-with-repo", the title is "Microservices without repo" and the name is also "without repo".

We need two cases:
1. Continuous deployment in existing repo
2. Continuous deployment in new repo

I am not sure if we need to call out microservices specifically. Customers can deploy multiple microservices in a single shot. I would say app deployment, and param name is "app name".

elamaran11: This is fixed. I dont understand the `cognitive dissonance` phrase because its a typo. Lets talk about these feedback,

PR #61, opened 2024-10-10
Interactive changes with @shapriov103

5herlocked:
*Issue #, if available:*

*Description of changes:*


applications/rust/buildspec.yaml
```diff
@@ -8,11 +8,11 @@ phases:
...
     commands:
       - cd ./applications/rust/
-      - docker build -t $ECR_REGISTRY/app/rust-microservice:$CODEBUILD_RESOLVED_SOURCE_VERSION .
-      - docker push $ECR_REGISTRY/app/rust-microservice:$CODEBUILD_RESOLVED_SOURCE_VERSION
+      - docker build -t $ECR_REGISTRY/app/rust-svc:$CODEBUILD_RESOLVED_SOURCE_VERSION .
+      - docker push $ECR_REGISTRY/app/rust-svc:$CODEBUILD_RESOLVED_SOURCE_VERSION
       - cd integration
-      - docker build -f Dockerfile -t $ECR_REGISTRY/$ECR_REPOSITORY:$CODEBUILD_RESOLVED_SOURCE_VERSION .
-      - docker push $ECR_REGISTRY/$ECR_REPOSITORY:$CODEBUILD_RESOLVED_SOURCE_VERSION
+      - docker build -f Dockerfile -t $ECR_REGISTRY/test/rust-svc:$CODEBUILD_RESOLVED_SOURCE_VERSION .
```
shapirov103: let's use the following nomenclature consistently for all apps:

$ECR_REGISTRY/moderneng/rust-microservice

repo is moderneng (prefix)

applications/next-js/pnpm-lock.yaml
```diff
@@ -0,0 +1,11142 @@
+lockfileVersion: '6.0'
```
shapirov103, line 1: Do we need this file in the repo or should it be in gitignore?

applications/next-js/lib/dynamo/index.ts
```diff
@@ -1,79 +1,79 @@
-import {v4 as uuidv4} from "uuid";
-import {Menu, Page, Product, Category, Cart, CartProduct} from "./types";
+import { v4 as uuidv4 } from 'uuid';
```
shapirov103, line 1: What is the reason to have direct access to dynamo from this app? This seems to violate the architecture - ui talks to the backend, backend (rust) talks to the db. 
With this architecture the service account will have to be created for both. It also creates lack of encapsulation - let's say rust moves to a different backend, the ui app will be affected as well. 
Let's discuss, I can defer this tech debt, but customers will see it. Maybe something that Qdev can help solve - let's think.

applications/next-js/.gitea/workflows/build.yaml
```diff
@@ -0,0 +1,31 @@
+name: NextJS App Build
```
shapirov103: Does it work? This looks really cool btw if it was validated.

applications/rust/deployment/local/otel-config.yaml
```diff
@@ -0,0 +1,20 @@
+receivers:
```
shapirov103: Do we need this file here? I assume it must be applied in the platform. If it is needed for more than just an example, please explain.

applications/rust/deployment/oam/rust-app-oam.yaml
```diff
@@ -1,51 +1,64 @@
...
-            allow:
-              - read
-              - write
-              - create
-        - type: cdn
+            domain: "*.amazonaws.com"
+            rewritePath: "/rust-app"
+            http:
+              /rust-app: 8080
+        - type: aws-service-account
```
shapirov103: Do we need this file for deployment? Compared to this one: https://github.com/aws-samples/appmod-blueprints/blob/main/deployment/teams/team-r/dev/application.yml
 that I demoed it seems redundant and won't work in the latest platform version. The deployment part will be handled during the workshop I assume and we will use consistent folder structure in the apps to place such artifacts.

applications/next-js/pnpm-lock.yaml
5herlocked: Requried for production builds of the website.

applications/next-js/lib/dynamo/index.ts
5herlocked: There is no direct access to dynamo from this app, the file itself calls API's behind the scenes hosted by the rust app.

It's the delineation of client side rendering vs. server side rendering

applications/rust/deployment/local/otel-config.yaml
5herlocked: Not required, it's something that I forgot to prune. The entire local folder should be gone.

applications/next-js/.gitea/workflows/build.yaml
5herlocked: Have not validated if it works, but gitea actions states it's fully compatible with GH Actions, so it _should_ work?

applications/rust/deployment/oam/rust-app-oam.yaml
5herlocked: Correct, this is a hold over from attempted isolated testing, will prune.

applications/next-js/.gitea/workflows/build.yaml
shapirov103: my concern is docker in docker setup inside the kubernetes cluster (mgmt EKS) - running this action requires docker env.

applications/next-js/lib/dynamo/index.ts
shapirov103: Oh that is better than I thought. let's rename this folder to something that is more descriptive for the backend API interaction.

applications/next-js/pnpm-lock.yaml
shapirov103: Are you saying it won't be generated automatically or are you pointing that it should be checked in for production reproducible builds? We can leave it, but these libs go stale probably faster than we can keep up to maintain.

applications/next-js/.gitea/workflows/build.yaml
shapirov103: And let's stick with DoD for PR - tested stuff only. I am happy to help you setup the platform locally to give it a shot. My suggestion is to wait for the platform to apply changes (due to this repo going public), then reprovision and test. If it works, it can change how we approach CI.

applications/next-js/.gitea/workflows/build.yaml
```diff
@@ -0,0 +1,31 @@
...
+
+jobs:
+  build-and-push:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+      - name: Login to Amazon ECR
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
```
shapirov103: How did you envision these credentials will be made available in GitEa?

5herlocked: Either Pod Identity Provider to give it assume role capabilities OR K8s secrets so we can retrieve at build time?

shapirov103: if it is clear which service account is used that needs pod identity (provided it is propagated properly) then yeah. As far as K8S secrets - what secret should be created? How do you associate a secret with the gitea action? I assume there is an approach, but all of that will be discovered once you test it out.

applications/rust/deployment/local/otel-config.yaml
shapirov103: Let's drop it then.

shapirov103 (approved):
LGTM, great work!

PR #64, opened 2024-10-16
OAM service account component, health checks, e2e rust backend

shapirov103:
*Issue #, if available:*

*Description of changes:*
Working implementation of service account, dependency management, health checks, example application deployment for rust with dynamo and pod identity

### DIFF
--- a/applications/rust/build.sh
+++ b/applications/rust/build.sh
@@ -1,6 +1,6 @@
 AWS_REGION=us-west-2
 CODEBUILD_RESOLVED_SOURCE_VERSION=latest
-AWS_ACCOUNT_ID=929819487611
+AWS_ACCOUNT_ID=$(aws sts get-caller-identity --output text --query Account)
 ECR_REGISTRY=$AWS_ACCOUNT_ID.dkr.ecr.$AWS_REGION.amazonaws.com
 ECR_REPOSITORY=moderneng/rust-microservice
 aws ecr get-login-password --region $AWS_REGION | docker login --username AWS --password-stdin $ECR_REGISTRY
--- a/deployment/teams/team-n/dev/application.yml
+++ b/deployment/teams/team-n/dev/application.yml
@@ -28,3 +28,15 @@ spec:
             http:
               /northwind-app: 80
       
+        - type: storage
+          properties:
+            secret:
+              - name: northwinde-dbinstance-out
+                mountPath: /opt/secret-volume
+  policies: # adopting secret that was generated outside of the app as readonly
+    - type: read-only
+      name: read-only
+      properties:
+        rules:
+          - selector:
+              resourceTypes: ["Secret"]
--- a/deployment/teams/team-n/dev/sqlserver-instance.yaml
+++ b/deployment/teams/team-n/dev/sqlserver-instance.yaml
@@ -21,7 +21,7 @@ spec:
       key: password
       name: northwind-dbinstance
       namespace: team-n
-    publiclyAccessible: false
+    publiclyAccessible: true
     region: us-west-2
     skipFinalSnapshot: true
     username: adminuser


deployment/teams/team-r/dev/application.yml
```diff
@@ -0,0 +1,53 @@
+apiVersion: core.oam.dev/v1beta1
```
shapirov103, line 1: @5herlocked @elamaran11  - here is an example of e2e rust backend deployment.

platform/traits/service-account.cue
```diff
@@ -1,93 +0,0 @@
-"aws-service-account": {
-	type: "trait"
```
elamaran11, line 2: Is this no longer needed?

applications/rust/build.sh
```diff
@@ -0,0 +1,11 @@
+AWS_REGION=us-west-2
+CODEBUILD_RESOLVED_SOURCE_VERSION=latest
+AWS_ACCOUNT_ID=929819487611
```
elamaran11: Remove account numbers

deployment/addons/kubevela/templates/components/ddbtable-component.yml
```diff
@@ -0,0 +1,63 @@
...
+  annotations:
+    definition.oam.dev/alias: ""
+    definition.oam.dev/description: Amazon DynamoDB Table
+  name: dynamodb-table
+spec:
+  schematic:
+    cue:
+      template: |
+        output: {
+        	apiVersion: "dynamodb.aws.upbound.io/v1beta1"
```
elamaran11, line 13: Does this work fine when abstracted?

platform/crossplane/compositions/dynamodb/ddb-table.yml
```diff
@@ -0,0 +1,82 @@
+apiVersion: apiextensions.crossplane.io/v1
+kind: Composition
```
elamaran11, line 2: Is this extension required? Default composition wont work. Are we discussing about this in slack?

platform/traits/service-account.cue
shapirov103: Yes, changed from trait to component otherwise race condition happens and I have no control.

applications/rust/build.sh
shapirov103: Oops.

deployment/addons/kubevela/templates/components/ddbtable-component.yml
shapirov103: works for any table any keys. It is limited atm as far as the breadth of DDB capabilities, but it is generic.

platform/crossplane/compositions/dynamodb/ddb-table.yml
shapirov103: This composition is the initial work that I am leading to improve our solution. Basic managed resources won't work as they require that developers pass cluster name and region. this is a clear security vulnerability plus it requires developers to know details that they are not supposed to know. My research show it can only be achieved by using EnvironmentConfig on crossplan which is not even enabled by default.

applications/rust/build.sh
shapirov103: addressed

platform/crossplane/compositions/dynamodb/ddb-table.yml
elamaran11: @shapirov103`environmentconfigs.apiextensions.crossplane.io` CRD is enabled in all mgmt and child clusters. So do you think you still have a problem with this. Plus since you are stating it has security vulnerability, do you think you still want to merge and iterate this or you want to remove this now? Im good either way. We want a clear design and followup approach before we can merge this.

shapirov103: CRD is enabled but I don't believe it will be honored in the actual composition. 
The actual deployment of crossplane must be modifed to include --enableEnvironmentConfig flag

```
$ kubectl edit deployment crossplane --namespace crossplane-system
apiVersion: apps/v1
kind: Deployment
spec:
# Removed for brevity
  template:
    spec:
      containers:
      - args:
        - core
        - start
        - --enable-environment-configs   
```

it can be done through helm from what I see, by overriding the args value.

elamaran11: I see, do you mind quickly modifying the helm release in this PR for all 3 clusters. If not we can do it separate

shapirov103: This is not relied on in this PR yet, so I am deferring to later stages. I will first create a composition that actually works with env config, then fix the platform to automate it. And enabling crossplane env conf is only part of it. Each cluster should actually have a valid EnvironmentConfig resource with the env. That maybe part of argo though.

elamaran11: Got it.

PR #70, opened 2024-10-23
Added env config to Crossplane in child clusters

hmuthusamy:
*Issue #, if available:*

*Description of changes:*
Crossplane config changes for child cluster

### DIFF
--- a/platform/infra/terraform/deploy-apps/manifests/crossplane-dev.yaml
+++ b/platform/infra/terraform/deploy-apps/manifests/crossplane-dev.yaml
@@ -11,6 +11,9 @@ spec:
     targetRevision: 1.17.1
     helm:
       releaseName: crossplane
+      values: |
+        args:
+          - --enable-environment-configs
   destination:
     name: 'dev-cluster'
     namespace: crossplane-system
--- a/platform/infra/terraform/setup-environments.sh
+++ b/platform/infra/terraform/setup-environments.sh
@@ -259,10 +259,12 @@ sleep 120
 # Setup CrossPlane IRSA for DEV Cluster
 aws eks --region $TF_VAR_aws_region update-kubeconfig --name $TF_VAR_dev_cluster_name
 kubectl apply -f ${REPO_ROOT}/platform/infra/terraform/deploy-apps/drc/crossplane-aws-drc-dev.yml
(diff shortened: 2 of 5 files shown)


elamaran11:
@hmuthusamy Why is this required? Is this a bug?

platform/infra/terraform/deploy-apps/manifests/crossplane-dev.yaml
```diff
@@ -11,6 +11,9 @@ spec:
     targetRevision: 1.17.1
     helm:
       releaseName: crossplane
+      values: |
```
shapirov103, line 14: are these additional args, or these will replace the args entirely? Because if the latter crossplane won't work.

platform/infra/terraform/setup-environments.sh
```diff
@@ -259,10 +259,12 @@ sleep 120
 # Setup CrossPlane IRSA for DEV Cluster
 aws eks --region $TF_VAR_aws_region update-kubeconfig --name $TF_VAR_dev_cluster_name
 kubectl apply -f ${REPO_ROOT}/platform/infra/terraform/deploy-apps/drc/crossplane-aws-drc-dev.yml
+kubectl apply -f ${REPO_ROOT}/platform/infra/terraform/deploy-apps/drc/cp-dev-env-config.yaml
```
shapirov103, line 262: We need a location where we can just throw a manifest without impact to the script. We can defer, it is just a callout. 

It looks like for terraform the actual IaC is the script not just the template.

shapirov103 (commented):
Please my question below.

platform/infra/terraform/deploy-apps/manifests/crossplane-dev.yaml
hmuthusamy: No. this is only to add additional args as per the documentation - https://docs.crossplane.io/latest/software/install/#feature-flags. we can add as many feature flags as needed. Checked the pods after testing locally as well

platform/infra/terraform/setup-environments.sh
hmuthusamy: We have a location already. If you have an argocd application that needs to be deployed in child cluster, add it to deploy-apps/manifests/ folder. Like we did for DAPR. the reason for having this in DRC is because of race conditions as Crossplane needs to be setup first before we install the components on child

shapirov103 (approved):
LGTM

PR #80, opened 2024-10-29
Adding metrics driven deployment into the manifest, with docs to expl…

Eli1123:
Main part works well. Missing a few parts, and needs some discussion. Enough for a general review to begin though.

Added a doc with extensive rundown of my additions and how to use it. Added a load test for the current application example since its needed to make the application emit metrics to track. Added changes to the CUE file, and it has another commented out option I'd like to discuss if it would be useful called evaluationCriteriaRatios in application.yml as well.

Existing issues:
- AMP URL is hardcoded currently. Ran into issues since YAML's address: "{{args.amp-workspace}}" does not work and cannot find a way to convert this to CUE currently.

### DIFF
--- a/deployment/teams/team-r/dev/application.yml
+++ b/deployment/teams/team-r/dev/application.yml
@@ -33,74 +33,61 @@ spec:
         targetPort: 8080
         replicas: 5
         serviceAccount: "rust-service-account"
-        dummyTestVariable: "test-67"
+        dummyTestVariable: "test-#" # (Optional) If changed and applied changes it can trigger a rollout for testing.
         functionalMetric:
-          interval: "1s" # user needs to add s, m, or h next to the number.
-          count: 1
           evaluationCriteria: [
             {
-              function: "sum",
+              interval: "1s", # user needs to add s, m, or h next to the number (second, minute, hour).
+              count: 1,
+              function: "sum", # Optional from list of "sum" | "avg" | "max" | "min" | "count"
+              successOrFailCondition: "success", # "success" or "fail" as options.
+              metric: "rocket_http_requests_total", # Required and is case and format sensitive.
+              comparisonType: ">", # Options ">" | ">=" | "<" | "<=" | "==" | "!="
+              threshold: 0
+            },
+            {
+              interval: "1s", 
               successOrFailCondition: "success",
-              metric: "rocket_http_requests_total",
+              function: "avg",
+              metric: "rocket_http_requests_duration_seconds_sum",
               comparisonType: ">",
               threshold: 0
             },
             {
+              interval: "1s", 
+              count: 1,
               successOrFailCondition: "success",
-              metric: "rocket_http_requests_total",
+              function: "max",
+              metric: "rocket_http_requests_duration_seconds_count",
               comparisonType: ">",
               threshold: 0
             },
             {
+              interval: "1s", 
+              count: 1,
+              function: "count",
               successOrFailCondition: "success",
-              metric: "rocket_http_requests_total",
+              metric: "rocket_http_requests_duration_seconds_bucket",
               comparisonType: ">",
               threshold: 0
             },
             {
-              successOrFailCondition: "success",
+              interval: "1s", 
+              count: 1,
+              successOrFailCondition: "fail",
               metric: "rocket_http_requests_total",
-              comparisonType: ">",
-              threshold: 0
-            },
-            {
-              successOrFailCondition: "success",
-              metric: "rocket_http_requests_total",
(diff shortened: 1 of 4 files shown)


deployment/teams/team-r/dev/application.yml
```diff
@@ -27,16 +27,76 @@ spec:
...
         port: 80
         targetPort: 8080
-        replicas: 1
+        replicas: 5
         serviceAccount: "rust-service-account"
-        # functionalGate:
-        #   pause: "10s" 
-        #   image: "public.ecr.aws/i8e1q7x5/appmod-javafunctest:latest"
-        #   extraArgs: "red"
+        dummyTestVariable: "test-67"
```
shapirov103: dummyTestVariable is a leftover?

deployment/teams/team-r/dev/application.yml
```diff
@@ -27,16 +27,76 @@ spec:
...
+          evaluationCriteria: [
+            {
+              function: "sum",
+              successOrFailCondition: "success",
+              metric: "rocket_http_requests_total",
+              comparisonType: ">",
+              threshold: 0
+            },
+            {
+              successOrFailCondition: "success",
```
shapirov103, line 50: what is the default function here? count?

deployment/teams/team-r/dev/application.yml
```diff
@@ -27,16 +27,76 @@ spec:
...
+              metric: "rocket_http_requests_total",
+              comparisonType: ">",
+              threshold: 0
+            },
+            {
+              successOrFailCondition: "success",
+              metric: "rocket_http_requests_total",
+              comparisonType: ">",
```
shapirov103, line 62: is there a default comparison type?

deployment/teams/team-r/dev/application.yml
```diff
@@ -27,16 +27,76 @@ spec:
...
-        replicas: 1
+        replicas: 5
         serviceAccount: "rust-service-account"
-        # functionalGate:
-        #   pause: "10s" 
-        #   image: "public.ecr.aws/i8e1q7x5/appmod-javafunctest:latest"
-        #   extraArgs: "red"
+        dummyTestVariable: "test-67"
+        functionalMetric:
+          interval: "1s" # user needs to add s, m, or h next to the number.
```
shapirov103: Is interval single and fixed?

deployment/teams/team-r/dev/application.yml
```diff
@@ -27,16 +27,76 @@ spec:
...
+          count: 1
```
shapirov103: Does count apply to all evaluation criteria?

shapirov103 (commented):
Please see some questions

deployment/teams/team-r/dev/application.yml
Eli1123: I feel it would be useful to keep as an option for when a user wants to trigger a rollout without actually changing anything important. But probably should not leave as 67 if you agree.

deployment/teams/team-r/dev/application.yml
Eli1123: I did have it like that at first but would probably be beneficial to let customers choose this for each one.

deployment/teams/team-r/dev/application.yml
Eli1123: In appmod-service.cue I have function as optional so it does not have default.

deployment/teams/team-r/dev/application.yml
Eli1123: Good point overlooked that. Making it `*">" | ...`

platform/components/appmod-service.cue
```diff
@@ -133,6 +154,50 @@ template: {
...
+								address: "https://aps-workspaces.us-west-2.amazonaws.com/workspaces/ws-6ba7e445-e203-43a5-b1b0-c7bc15e354ef"
+								query: [
+									if criteria.function != _|_ {
+										"\(criteria.function)(\(criteria.metric))"
+									}
+									if criteria.function == _|_ {
+										"\(criteria.metric)"
+									}
+								][0]
+								authentication: sigv4: region: "us-west-2"
```
shapirov103: this part is something that makes it a POC rather than more production oriented. What if the region is different? Many customers work of different regions, how will that be handled? Can you check if there is a way to have region set on the service account or somewhere where it does not make into the component definition? 

Of course, one approach is to collect the region as a param from the customer.

Eli1123: Great point, overlooked that hard coded region

PR #82, opened 2024-10-29
Feature/db rds ec2 changes

hmuthusamy:
*Issue #, if available:*

*Description of changes:*

1. Added timeout for LB tf
2. Updated sleeptime and improved script to run little faster on the re-run
3. Added Godwin and Ramesh Database changes and cleaned up some the code.


packages/backstage/dev/cm-backstage-config.yaml
```diff
@@ -112,7 +112,7 @@ data:
       locations:
         # Examples from a public GitHub repository.
         - type: url
-          target: https://github.com/aws-samples/appmod-blueprints/blob/main/platform/backstage/templates/catalog-info.yaml
+          target: https://github.com/aws-samples/appmod-blueprints/blob/feature/db-rds-ec2-changes/platform/backstage/templates/catalog-info.yaml
```
elamaran11: this wont work even for your testing. You need absolute main HEAD URL for backstage to work.

elamaran11 (commented):
See the comment

platform/infra/terraform/database/README.md
```diff
@@ -0,0 +1,122 @@
...
+- EC2 instance with SQL Server
+- Security group for the EC2 instance
+- AWS Secrets Manager secret for storing credentials
+- SSH key pair for instance access
+
+The EC2 instance is configured with:
+- SQL Server
+- AWS CLI
+- EC2 Instance Connect
+- Java Runtime Environment (JRE)
```
elamaran11, line 17: Why is this required?

platform/infra/terraform/setup-environments.sh
```diff
@@ -110,7 +113,7 @@ echo "Github Source Repo for Management EKS Cluster =" $GITHUB_URL
 cd "${REPO_ROOT}/platform/infra/terraform/"
 
 rm -rf terraform-aws-observability-accelerator/
-git clone https://github.com/aws-observability/terraform-aws-observability-accelerator.git
+git clone https://github.com/hmuthusamy/terraform-aws-observability-accelerator.git -b feature/enable-polyglot-otel-config
```
elamaran11, line 116: Are we not replacing this to use the upstream repo. Lets use the collector config method proposed by Rodrigue and use upstream

elamaran11 (commented):
@hmuthusamy This is fanstastic work. Only feedback i have is, we should not use fork. We should think about using the upstream with using collector config.

platform/infra/terraform/setup-environments.sh
hmuthusamy: We have to. But we will have to evaluate if this will work with all languages and check the configs. This fork repo has been tested out locally multiple times and opens up the OTEL config. We can use upstream repo when our config is ready and tested

elamaran11: No my point is different. We dont need enable logic like you. All we need is a custom collector config added on top of upstream repo and then run teraform accelerator. Lets try to do that as we discussed.

elamaran11 (approved):
LGTM but we need to fix the Collector config and remove your fork.

PR #87, opened 2024-10-31
Parallel Run, remove prod-dev clustser in cmbackstage and cleanup

hmuthusamy:
*Issue #, if available:*

*Description of changes:*

1. Removed README from database folder
2. Removed dev and prod cluster reference in the CM config for backstage
3. Added Wait and parallel terraform apply to reduce the time


packages/backstage/dev/cm-backstage-config.yaml
```diff
@@ -135,13 +135,5 @@ data:
               url: http://${DOMAIN_NAME}/argocd
               username: admin
               password: ${ARGOCD_ADMIN_PASSWORD}
-            - name: dev-cluster
```
shapirov103, line 138: Why is this not necessary anymore after this change?

shapirov103 (commented):
@hmuthusamy LGTM, please see my small question. 

Was it tested? (assume it was)

hmuthusamy:
> @hmuthusamy LGTM, please see my small question.
> 
> Was it tested? (assume it was)

Yes. Tested multiple times locally destroyed and recreated. Also re-run was quicker.

packages/backstage/dev/cm-backstage-config.yaml
hmuthusamy: We are deploying the templates only backstage in mgmt cluster. DEV and PROD cluster dont have argocd installed. So when deploying templates you see 3 iterations instead of one right one. This will fix that issue

shapirov103: makes sense

platform/infra/terraform/database/README.md
```diff
@@ -1,122 +0,0 @@
-# Terraform AWS Infrastructure Project
-
```
elamaran11: Why is this file removed. I think it can exist without harm. Its to say how the script works.

elamaran11 (commented):
One Minor comment.

platform/infra/terraform/database/README.md
hmuthusamy: @elamaran11 Added it back

elamaran11 (approved):
LGTM

PR #130, opened 2024-11-12
Reverting kubevela change for metrics driven decision

shapirov103:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/deployment/addons/kubevela/templates/components/appmod-service.yaml
+++ b/deployment/addons/kubevela/templates/components/appmod-service.yaml
@@ -10,9 +10,6 @@ spec:
     cue:
       template: |2
         let previewService = "\(context.name)-preview"
-        let ampWorkspaceUrl = "{{args.amp-workspace-url}}"
-        let ampWorkspaceRegion = "{{args.amp-workspace-region}}"
-        let prometheusTargetQuery = "k8s_container_name=\"\(parameter.image_name)\", k8s_namespace_name=\"\(context.namespace)\""
 
         output: {
         	apiVersion: "argoproj.io/v1alpha1"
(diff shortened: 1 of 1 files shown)


hmuthusamy (approved):

deployment/addons/kubevela/templates/components/appmod-service.yaml
```diff
@@ -318,7 +201,6 @@ spec:
         	serviceAccount:   *"default" | string
```
elamaran11, line 201: Why is this parameter here if not used?

deployment/addons/kubevela/templates/components/appmod-service.yaml
```diff
@@ -295,20 +192,6 @@ spec:
...
-        		function?:              "sum" | "avg" | "max" | "min" | "count"
-        		successOrFailCondition: *"success" | "fail"
-        		metric:                 string
-        		comparisonType:         *">" | ">=" | "<" | "<=" | "==" | "!="
-        		threshold:              *0 | number
-        	}]
-        }
-
         parameter: {
         	image_name:       string
```
elamaran11, line 196: Why is this parameter here if not used?

elamaran11 (commented):
@shapirov103 Have some comments and questions. I see whole of the AppMod deployment functionality is removed so you going to resue this for functional tests too. If so can you rename the file accordingly? Also have two things to fix!

deployment/addons/kubevela/templates/components/appmod-service.yaml
shapirov103: which parameter is not used? parameter image_name is used when this component is applied with application.yml

PR #139, opened 2024-11-12
add image update step to cache warm build

zjaco13:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/platform/backstage/templates/cicd-pipeline/provisioner/wf-templates.yaml
+++ b/platform/backstage/templates/cicd-pipeline/provisioner/wf-templates.yaml
@@ -186,6 +186,20 @@ spec:
                 value: cache-init
           depends: "clone"
 
+        - name: edit-tag
+          template: update-image-tag
+          arguments:
+            parameters:
+              - name: env
+                value: dev
+              - name: tag
+                value: "{{workflow.parameters.tag-version}}"
+              - name: apprepo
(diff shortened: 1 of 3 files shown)


platform/backstage/templates/cicd-pipeline/provisioner/wf-templates.yaml
```diff
@@ -266,5 +281,51 @@ spec:
                 --skip-tls-verify"
                 ]
 
+  - name: update-image-tag
```
shapirov103, line 284: copy pasted?

shapirov103 (commented):
Thanks for the quick turn around. curious if this workflow allows any reuse mechanism. I assume it is a copy paste, but curious if the workflows become more complex whether we can reuse steps between workshops.

platform/backstage/templates/cicd-pipeline/provisioner/wf-templates.yaml
zjaco13: yes it was.  I already have this cache warmer as a template, but now that its the same as the cicd workflow I can probably reuse there

zjaco13: will look into

shapirov103 (approved):
LGTM

PR #148, opened 2024-11-15
Cleaned up for workshop

5herlocked:
*Issue #, if available:*

*Description of changes:*


shapirov103 (commented):
Why is rust dashboard deleted? I thought we could use it in the future to show how dev teams can create their own dashboards.

5herlocked:
application yml will be moved to the reference folder,

I can add the dashboards back in however the non OAM manifests were removed to reduce the confusion, though they can be moved to reference like a: "this is all that you have to do without OAM"

shapirov103:
> application yml will be moved to the reference folder, and there's a reference manifest included in the WS instructions
> 
> I can add the dashboards back in however the non OAM manifests were removed to reduce the confusion, though they can be moved to reference like a: "this is all that you have to do without OAM"

Good on app.yml - sounds like a plan. 

Let's keep the dashboard in the reference folder.

shapirov103 (approved):
LGTM

PR #171, opened 2024-12-18
Feature/backstage sql spl chars

hmuthusamy:
*Issue #, if available:*

*Description of changes:*
Removed special characters on the backstage pwd

### DIFF
--- a/platform/infra/terraform/mgmt/terraform/backstage.tf
+++ b/platform/infra/terraform/mgmt/terraform/backstage.tf
@@ -1,7 +1,7 @@
 resource "random_password" "backstage_postgres_password" {
   length           = 48
-  special          = true
-  override_special = "@$"
+  special          = false
+  # override_special = "!#"
 }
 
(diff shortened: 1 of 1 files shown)


platform/infra/terraform/mgmt/terraform/backstage.tf
```diff
@@ -1,7 +1,7 @@
 resource "random_password" "backstage_postgres_password" {
   length           = 48
```
shapirov103: length is crazy to me as well. I see some folks had issues when it was just truncated. Let's make it 16 or something like that

hmuthusamy: Fixed it to 16

elamaran11: I agree

elamaran11: Lets test it

elamaran11 (approved):

Issue #180, opened 2025-02-19
error external secret duplicate when using componentdefinition appmod-service


csantanapr:
When creating two services with appmod-server componentdefinition appmod-service https://github.com/aws-samples/appmod-blueprints/blob/main/deployment/addons/kubevela/templates/components/appmod-service.yaml 

It creates duplicate resources for external secret `amp-workspace-secrets`
```
"amp-workspace-secrets": {
        		apiVersion: "external-secrets.io/v1beta1"
        		kind:       "ExternalSecret"
        		metadata: {
        			name:      "amp-workspace-secrets"
        			namespace: context.namespace
```
you get the error when creating the second applications using appmod-service
```
    - firstExecuteTime: "2025-02-19T05:09:28Z"
      id: lmn7ysq2ke
      lastExecuteTime: "2025-02-19T05:10:20Z"
      message: 'run step(provider=oam,do=component-apply): Dispatch: pre-dispatch
        dryrun failed: Found 1 errors. [(cannot apply ApplyOption: existing object
        ExternalSecret default/amp-workspace-secrets is managed by other application
        default/next-js-application)]'
      name: progressive-demo
```

shapirov103:
@Eli1123  please take a look when you get a chance. I recall Zach taking over this part, however I see in the original PR the external secret name is hardcoded, so deployments to the same namespace are destined to fail. 
Is there a way to make this secret app specific (e.g. using context.name)?
@zjaco13 if you recall please add your perspective as well.

zjaco13:
Ack, see the issue, can work on it

elamaran11:
Fixed

PR #181, opened 2025-02-20
hotfix for deploying multiple apps

zjaco13:
*Issue #, if available:*
#180
*Description of changes:*

Hotfix for deploying multiple applications into the same namespace.  Adds the component name to the end of the secret.

Need to look into kubevela shared resources(https://kubevela.io/docs/end-user/policies/shared-resource/) to see if fix is possible using this/if it would be a better approach


csantanapr (approved):
Thanks for the fix!

elamaran11 (approved):
LGTM

Issue #183, opened 2025-02-24
Add ACK content


csantanapr:

allamand:
kro/ack and kubevela and cross plane are supported now.

We default everything to kro/ACK

PR #184, opened 2025-02-24
Add support for ACK to the Platform Engineering workshop

csantanapr:
*Issue  #, if available:*
#183 

*Description of changes:*
Adds example of s3 bucket with kubevela and backstage


.devcontainer.json
```diff
@@ -0,0 +1,9 @@
+{
+    "image":"mcr.microsoft.com/devcontainers/base:ubuntu",
+    "features": {
+    "ghcr.io/devcontainers/features/docker-in-docker:2": {},
+    "ghcr.io/devcontainers-extra/features/kind:1": {},
```
elamaran11, line 5: Why do we need DIND and Kind. Our setup is EKS implmentation isnt it? Having `devContainer` with terraform and all other pre reqs of workshop content like terraform etc would be ideal!

platform/backstage/templates/catalog-info.yaml
```diff
@@ -6,7 +6,8 @@ metadata:
 spec:
   targets:
     - ./rds-cluster/template.yaml
-    - ./s3-bucket/template.yaml
+    #- ./s3-bucket/template.yaml
```
elamaran11: remove

elamaran11 (commented):
@csantanapr Everything looks great except few minor comments. We can merge this only after ACK install PR changes are merged!

shapirov103 (commented):
Requires functional validation after ACK is added to the provisioning.

zjaco13:
Do we need to keep the crossplane s3 bucket and backstage template?  Or can we get rid of it in this PR as well?

elamaran11:
> Do we need to keep the crossplane s3 bucket and backstage template? Or can we get rid of it in this PR as well?

We should replace it.

.devcontainer.json
csantanapr: I use kind on ec2 via devpod, this allows me to use the instance IAM Role allow ACK/Crossplane and any pod that needs IAM authentication to work. 
This made easy to unit test

csantanapr: the current devcontainer doesn't create a kind cluster, just installs the kind CLI

platform/backstage/templates/catalog-info.yaml
csantanapr: removed

.devcontainer.json
csantanapr: added terraform cli

zjaco13:
Tested successfully with ACK Changes

shapirov103:
@csantanapr if there are any regression issues for the workshop given this change, let's change the target from `main` to `dev`. `dev` branch should be used for all the changes to avoid regression. 
We can adjust it if needed, but given the current state of the workshop with all the github references in scripts and Argo apps, this seems like the safest route for now.

csantanapr:
closing this one as we are pushing to dev first

PR #200, opened 2025-03-17
Merging changes from dev to main

mmaroth:
*Issue #, if available:*  Extended CICD scaffolder template to include argocd and argo workflow details in component overview and cicd tab.

*Description of changes:* We have tested this change in TE last week, and reviewed with Carlos.


elamaran11 (commented):
LGTM. The merging will blocked unless a peer run is done on WS Studio pointing to Dev branch.

elamaran11:
LGTM

elamaran11 (approved):
LGTM

packages/backstage/dev/cm-backstage-config.yaml
```diff
@@ -112,7 +112,7 @@ data:
       locations:
         # Examples from a public GitHub repository.
         - type: url
-          target: https://github.com/aws-samples/appmod-blueprints/blob/main/platform/backstage/templates/catalog-info.yaml
+          target: https://${DOMAIN_NAME}/gitea/giteaAdmin/platform/src/branch/main/backstage/templates/catalog-info.yaml
```
shapirov103, line 115: backstage cm config moved to gitea from github 
Is that what we want? would it better to just tag the main on some release and point to that rather than make it dependant on gitea?
If we tag, it will prevent volatility, i believe that was the reason for this move.

csantanapr: We want it on gitea, to make the workshop more robust as it uses the copy of the code at the time the workshop is provisioned.

csantanapr: We need to stop referring back to github.com from workshops, and make everything more isolated, this way easier for customer to take solution and customize to use a git server of their own.

shapirov103: My concern was around spreading gitea in our configs which makes the solution repo specific to the workshop rather than being a solution repo. This can be addressed separately, especially if we plan to support github enterprise deployments. we can create a playbook of what needs to be updated, or parameterize the whole git URL.

PR #208, opened 2025-03-24
Backstage template updates

mmaroth:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/platform/backstage/templates/apigw-sqs-terraform/template-apigw-sqs-terraform.yaml
+++ b/platform/backstage/templates/apigw-sqs-terraform/template-apigw-sqs-terraform.yaml
@@ -22,10 +22,6 @@ spec:
             - name
             - aws_region
           type: object
-        repoUrl:
-          description: Gitea URL on management cluster
-          title: Gitea URL
-          type: string
         adminRoleName:
           description: Name of the role to give the administrative rights
           title: Admin role name
@@ -35,9 +31,12 @@ spec:
           title: Existing namespace to create this resource
           type: string
       title: Configuration options
-      required:
-        - repoUrl
   steps:
+    - id: fetchSystem
+      name: Fetch System
+      action: catalog:fetch
+      input:
+        entityRef: system:default/system-info
     - action: fetch:template
       id: fetch-base
       input:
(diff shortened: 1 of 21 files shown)


platform/backstage/templates/apigw-sqs-terraform/template-apigw-sqs-terraform.yaml
```diff
@@ -54,7 +53,7 @@ spec:
       input:
         description: Example of API Gateway with SQS # change this
         # Hard coded value for this demo purposes only.
-        repoUrl: ${{parameters.repoUrl}}?repo=${{parameters.tfVars.name}}
+        repoUrl: ${{ steps['fetchSystem'].output.entity.spec.hostname }}/gitea?repo=${{parameters.tfVars.name}}
```
elamaran11, line 56: How does this work? This causes cascading effect and we might need to make this change in to many modules before we publish this.

elamaran11 (commented):
@mmaroth This causes cascading effect and we might need to make this change in to many modules before we publish this. If you are up for it, i can publish this.

platform/backstage/templates/apigw-sqs-terraform/template-apigw-sqs-terraform.yaml
mmaroth: We are creating system entity with FQDN/hostname of gitea, as part of template we are fetching the value of it and passing to skelton.

csantanapr (approved):
This looks great @mmaroth thank you, this change will make all the templates consistent on hostname/dns

platform/backstage/templates/apigw-sqs-terraform/template-apigw-sqs-terraform.yaml
csantanapr: @elamaran11 the change is per template basis, take a look above the template has a step isolated the current template
```
  steps:
    - id: fetchSystem
      name: Fetch System
      action: catalog:fetch
      input:
        entityRef: system:default/system-info
```

csantanapr: @mmaroth already did the change in all other templates in this PR, to make it consistent across all the modules.

elamaran11: @csantanapr Agreed, i see he has made the change to all templates. This change also impacts all the module screenshots and content because we are asking people to input repoURL but with this change, thats no longer there. SO we want commitment to update all modules in workshop content, before i can merge it to Dev.

PR #230, opened 2025-04-17
setup environment config for crossplane on mgmt cluster

zjaco13:
Adds environment config for VPC and Subnets to enable crossplane compositions to use these values.
*Issue #, if available:*

*Description of changes:*


platform/infra/terraform/mgmt/terraform/scripts/crossplane/.gitignore
```diff
@@ -0,0 +1 @@
+*.yml
```
shapirov103: what is up with this entry? You don't want to include any yml files in the commits?

platform/infra/terraform/mgmt/terraform/scripts/crossplane/cp-mgmt-env-config.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: apiextensions.crossplane.io/v1alpha1
+kind: EnvironmentConfig
+metadata:
+  name: mgmt-environment
```
shapirov103, line 4: I assume we will be creating env configs for each environment, like dev/prod. In fact, the actual deliverable for env provisioning is populated env config. 

This VPC will only be used for things that are needed for the platform. If I need to pass VPC ID to a dynamo instance in dev, there should be a way to specify the env when we wire XRDs.

platform/infra/terraform/setup-environments.sh
```diff
@@ -55,11 +55,21 @@ sed -e "s/INGRESS_DNS/${DNS_HOSTNAME}/g" ${REPO_ROOT}/platform/infra/terraform/m
 export GITHUB_URL=$(yq '.repo_url' ${REPO_ROOT}/platform/infra/terraform/mgmt/setups/config.yaml)
 export GITHUB_BRANCH=$(yq '.repo_branch' ${REPO_ROOT}/platform/infra/terraform/mgmt/setups/config.yaml)
 
-
 # Deploy the apps on IDP Builder and ArgoCD
 ${REPO_ROOT}/platform/infra/terraform/mgmt/setups/install.sh
 
 cd ${REPO_ROOT}/platform/infra/terraform/mgmt/terraform/mgmt-cluster/
+
+VPC_ID=$(terraform output -raw eks_cluster_vpc_id)
```
shapirov103, line 64: we should find a better way to deal with this. Spreading this logic across the sh script is increasing our tech debt. 

What I would like to see from the functional perspective:

1. When mgmt cluster is stood up, the mgmt env config is created ootb by the IaC.
2. When any other env is created, such as dev/prod, at the point when provisioning is complete it must transactionally create env config for that env. transactional means that if provisioning failed, it should roll baclk. 

We can defer these till later, however I want to discuss it as a team and start taking steps to decrease the entropy.

shapirov103 (commented):
Looks great, I added a few comments.

platform/infra/terraform/mgmt/terraform/scripts/crossplane/.gitignore
zjaco13: When the values in the environment config get substituted from cp-mgmt-env-config.yaml, I move it to cp-mgmt-env-config.yml and then apply that one.  This ignore is just to make sure the populated values aren't accidentally pushed to git.

shapirov103: i suggest using a more explicit convention, e.g. calling the generated file as yaml.gen and excluding that one. Unless we have already a precedent that you followed.

platform/infra/terraform/mgmt/terraform/scripts/crossplane/cp-mgmt-env-config.yaml
zjaco13: Yes, there is a way to do this with EnvironmentConfig Selectors. I can add my example crd to this PR with that in it as well

platform/infra/terraform/mgmt/terraform/scripts/crossplane/.gitignore
zjaco13: There are a few other examples where its done like this in the repo.  I agree that it is probably an anti-pattern, so I'll update here

platform/infra/terraform/mgmt/terraform/scripts/crossplane/cp-mgmt-env-config.yaml
zjaco13: done

deployment/addons/kubevela/templates/components/rds-cluster-mysql.yaml
```diff
@@ -0,0 +1,49 @@
...
+        	apiVersion: "awsblueprints.io/v1alpha1"
+        	kind:       "RelationalDatabase"
+        	metadata: {
+        		name:      context.name
+        		namespace: context.namespace
+        	}
+        	spec: {
+        		compositionSelector: matchLabels: {
+        			"awsblueprints.io/provider":            "aws"
+        			"awsblueprints.io/environment":         "prod"
```
shapirov103, line 22: Prod hardcoded? We have selectors, i assume this should an input parameter, unless you have a clear path to avoid replication of such resources across each environment.

shapirov103 (commented):
I see more files were added to the review, potentially inadvertently. Added one comment.

elamaran11:
Lets wait for #235 to merge to main.

elamaran11 (changes requested):
Lets wait for https://github.com/aws-samples/appmod-blueprints/pull/234 to merge to main and then we can merge to Dev.

deployment/addons/kubevela/templates/components/rds-cluster-mysql.yaml
zjaco13: These files were added to show an example of what you can do with environment config.  This environment is hardcoded as a composition selector to make sure it chooses the myql one, the actual environment selection happens from the deploy parameter

shapirov103: I can see the source of confusion. If I see a component definition under addons/kubevela/templates I assume these are generic components, not just examples, that customers will be able to use across the board and I believe it is your intention - generic component.

A few questions:
1. what is the point of selecting the composition based on selectors as opposed to just using the name of the composition - e.g. compositionRef? Are you planning ahead for versioning?
2. Assuming there is a case to select composition based on selectors, what does `environment` represent as a label (other than being an example that many are using) and how it is related to the envs that we maintain - mgtm, dev, prod atm?
3. If this environment is orthogonal to the envs in our solution, then my suggestion is to get rid of this label to avoid confusion. 

In general, let's stick to the consistent naming, e.g. any reference to env or environment must be in the context of the solutions env. From that perspective, the `deploy` param should really be `env` and match to the env configs. 

If composition selectors are used for let's say versioning, then let's introduce versioning and use this as a pattern across the board.  

Happy to hear your thoughts on this if I am missing anything.

zjaco13: I copied the composition selector by labels from the other rds crossplane composition example we currently have in the repo.  I believe the meaning of the "environment" label is for full aurora clusters to be "prod", and just rds instances to be "dev" or "staging".  I see how that is confusing though, and agree that naming conventions should be changed.  This is intended for use as a generic component.

shapirov103: These files were added in May 2024 and I don't think they are used anywhere. Back then we did not have any environments yet. 
As far as labels - I don't mind having a very explicit claim like PostgresRDSInstance and AuroraPostgresRDSInstance when instantiating resources. 
In most cases this label selection is used for multi-cloud envs. However, i don't have a strong preference here, if you see a label matching case, we can keep it.

shapirov103:
@elamaran11 #234 is merged. let's push it.

elamaran11 (approved):
LGTM

PR #236, opened 2025-05-08
fix: UX issue when NextJS is started before Backend is ready

5herlocked:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/applications/next-js/Dockerfile
+++ b/applications/next-js/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:18-alpine AS base
+FROM node:20-alpine AS base
 
 # Install dependencies only when needed
 FROM base AS deps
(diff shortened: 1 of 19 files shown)


shapirov103 (approved):
LGTM - a lot of changes

shapirov103:
There is a bug in the kubevela ingress when URL rewrite is not performed. I will open a PR on that shortly.

5herlocked:
Just fully upgraded next-js itself which caused the lock file to get upgraded. Required for consistent deployments 😁

shapirov103:
> Just fully upgraded next-js itself which caused the lock file to get upgraded. Required for consistent deployments 😁

Thank you for that, ownership is a big part that keeps this solution alive.

PR #237, opened 2025-05-08
When regex is used and rewrite is set to false the trait spits out unroutable ingress

shapirov103:
*Issue #, if available:* based on user feedback from the delivery

*Description of changes:* allows regex shape for path, without rewrite of extended path. Affects UI application in module 2 for the rust application (at least).


elamaran11 (approved):
GTM

PR #242, opened 2025-06-03
Fixed sed command to work for Mac and Linux

mmaroth:
*Issue #, if available:*

*Description of changes:*


mmaroth:
The sed command is not working as expected when we run setup-env local, which is causing to not replace tokens, and its breaking backstage scaffolder templates. This issue happening only for local set up (MacOS). So fixing it to support both Linux and MacOS

PR #247, opened 2025-08-07
Adding Solution Tracking thru Open Search SAES Metrics

elamaran11:
*Issue #, if available:*

*Description of changes:*


shapirov103 (commented):
I have a concern about this identifier. The onboarding ticket states (uksb-u7h6yeonth) in round brackets. I think it is important to preserve them - based on what I see we did in EKS blueprints.

elamaran11:
@shapirov103 There is no concern here because if you see [this](https://github.com/aws-samples/appmod-blueprints/blob/fb990bd436d6518790f0863cd0a66ac9ccf3bb60/platform/infra/terraform/mgmt/terraform/mgmt-cluster/main.tf#L77) code, it introduces round brackets. If i had another round brackets in vars it will be two round brackets and will not track. I had same issue while testing and i have to remove it. Please check both commits.

PR #269, opened 2025-09-10
Feature/resource prefix gitops

allamand:
Fix usage of prefix_resource in gitops.

Add new script to register spoke clusters


hmuthusamy (approved):
LGTM

PR #279, opened 2025-09-15
updated overwrite settings

hmuthusamy:
*Issue #, if available:*

*Description of changes:*
more fixes for amp/amg

### DIFF
--- a/platform/infra/terraform/common/observability.tf
+++ b/platform/infra/terraform/common/observability.tf
@@ -13,7 +13,14 @@ locals {
 
 # Create the secret
 resource "aws_secretsmanager_secret" "argorollouts_secret" {
-  name = "/platform/amp"
+  name                    = "/peeks/amp"
+  recovery_window_in_days = 0
+
+  lifecycle {
+    ignore_changes = [name]
+  }
+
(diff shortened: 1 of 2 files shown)


platform/infra/terraform/common/observability.tf
```diff
@@ -13,7 +13,14 @@ locals {
 
 # Create the secret
 resource "aws_secretsmanager_secret" "argorollouts_secret" {
-  name = "/platform/amp"
+  name                    = "/peeks/amp"
```
allamand: this should use ${local.context_prefix}

allamand:
```suggestion
  name                    = "${local.context_prefix}-amp"
```

allamand: the secret name seems to be for argorollouts_secret, but the name has amp. Where this is read, that will need to be updated also

hmuthusamy: Changed. Yes. this is used by argo rollouts to send application data to AMP. the secret name will anyway needs to be changed along with cluster names which is next steps

PR #284, opened 2025-09-15
Feature/update backstage template

allamand:
update backstage templates


scripts/3-register-terraform-spoke-clusters.sh
```diff
@@ -124,6 +125,9 @@ externalSecret:
     enable_ack_ec2: "true"
     enable_ack_eks: "true"
     enable_ack_iam: "true"
+    enable_ack_s3: "true"
+    enable_kubevela: "true"
+    sebtest: "true"
```
hmuthusamy: Do we need this ? just checking to see if it was part of some testing

allamand: as for now, we create/update the cluster file with this script, but you are right, I think the value comes from the secret created by terraform spoke, and not the values here. need some tests to confirm, in such case we should be able to remove thoses. maybe create a following issues for this

platform/backstage/templates/create-dev-and-prod-env/template-create-dev-and-prod-env.yaml
```diff
@@ -46,7 +46,7 @@ spec:
           type: string
         repoHostUrl:
           type: string
-          default: <git_hostname>
+          default: d2jl8inhwf8wdv.cloudfront.net
```
hmuthusamy, line 49: seems like hardcoding?

allamand:
```suggestion
          default: <git_hostname>
```

hmuthusamy:
LGTM

PR #314, opened 2025-09-26
fix: add missing releaseName to crossplane addons for autosync

allamand:
## Problem
Crossplane and crossplane-aws ApplicationSets were missing automated sync policy while other addons like cert-manager had it correctly configured.

## Root Cause
The crossplane addon configurations in addons.yaml were missing the `releaseName` field, which caused the ApplicationSet template to process them differently than other addons.

## Solution
- Add `releaseName: crossplane` to crossplane addon configuration
- Add `releaseName: crossplane-aws` to crossplane-aws addon configuration

## Result
- Ensures ApplicationSet template processes crossplane addons consistently with other addons
- Resolves missing automated sync policy in generated ApplicationSets
- Crossplane applications will now have autosync enabled like other addons

## Testing
After this change, the generated ApplicationSets should include:
```yaml
syncPolicy:
  automated:
    allowEmpty: true
    prune: false
    selfHeal: false
```


hmuthusamy (approved):
LGTM

PR #356, opened 2025-10-22
rename yml to yaml & update kro to 0.5

allamand:
*Issue #, if available:*

*Description of changes:*


hmuthusamy (approved):

platform/backstage/customtemplates/ddb-table/template.yaml
```diff
@@ -63,7 +57,6 @@ spec:
       input:
         url: ./skeleton
         values:
-          provider: ${{ parameters.provider }}
```
punkwalker, line 66: Do we not need provider name now?

gitops/addons/default/addons/mlflow/values.yaml
```diff
@@ -1,113 +1,67 @@
...
-    auth:
-      enabled: true
-      type: oidc
-      oidc:
-        clientId: mlflow
-        scopes: 'openid profile email groups'
-        usernameKey: preferred_username
-        groupsKey: groups
+image:
+  repository: burakince/mlflow
```
punkwalker, line 5: Is this image needed just to move from bitnami?

allamand: not sure where this come from, mlflow still not working though

punkwalker: Let's ignore it then for now

platform/backstage/customtemplates/ddb-table/template.yaml
allamand: I don't think so, this remove it right ?

punkwalker: Okay, Just wanted to make sure template skeleton is rendered correctly

hmuthusamy (approved):

PR #370, opened 2025-10-30
Add Measuring Platform Success Requirements

zjaco13:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/gitops/addons/charts/crossplane-aws/templates/compositions.yaml
+++ b/gitops/addons/charts/crossplane-aws/templates/compositions.yaml
@@ -322,8 +322,10 @@ spec:
     - type: Selector
       selector:
         mode: Single
+        sortByFieldPath: metadata.name
         matchLabels:
-        - key: env
+        - fromFieldPathPolicy: Required
+          key: env
           type: FromCompositeFieldPath
           valueFromFieldPath: spec.deploy
   writeConnectionSecretsToNamespace: crossplane-system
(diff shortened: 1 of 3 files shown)


gitops/addons/bootstrap/default/addons.yaml
```diff
@@ -812,6 +812,78 @@ jupyterhub:
...
+        external:
+          url: "{{.metadata.annotations.aws_grafana_url}}"
+      mysql:
+        useExternal: true
+        externalServer: "devlake-mysql-service"
+        externalPort: "3306"
+      option:
+        database: mysql
+        connectionSecretName: "devlake-mysql-auth"
+        autoCreateSecret: false
```
allamand: We should keep values minimal in this file only ou values that needs to be dynamic from secret annotations. For the other ones we should put them in the default folders

platform/infra/terraform/common/secrets.tf
```diff
@@ -35,6 +35,33 @@ resource "random_password" "keycloak_admin" {
   }
 }
 
+resource "random_password" "devlake_encryption_secret" {
+  length  = 128
```
hmuthusamy: @zjaco13  do we need a long password?

gitops/addons/bootstrap/default/addons.yaml
zjaco13: That makes sense, I didn't see those values files used anywhere.

PR #409, opened 2025-12-03
Add automatic terraform lock handling to deploy and destroy scripts

allamand:
- Add proactive lock detection and clearing after terraform init
- Add reactive lock recovery on apply/destroy failures
- Apply to both common and cluster terraform stacks
- Prevents failures from stale locks left by incomplete operations

*Issue #, if available:*

*Description of changes:*

When cloudformation tried to recreate the stack, sometimes, it is not working due to terraform lock.
the goal of this change is to allow the script to force unloack before trying again


hmuthusamy (approved):
LGTM

PR #442, opened 2026-01-23
feat: Add EKS capabilities integration 

allamand:
- Add EKS cluster Terraform configuration updates
- Add Identity Center Terraform module for IAM integration
- Update deployment scripts with force-unlock helper
- Add hub-config.yaml changes for EKS capabilities
- Update ArgoCD and secrets Terraform configurations
- Add EKS Capabilities ArgoCD setup documentation

*Issue #, if available:*

*Description of changes:*


platform/infra/terraform/scripts/1-tools-urls.sh
```diff
@@ -50,8 +50,19 @@ pad_string() {
...
-ARGOCD_URL="https://$DOMAIN_NAME/argocd"
+# Get ArgoCD URL from EKS capability if available
+ARGOCD_SERVER_URL=$(aws eks describe-capability --cluster-name ${RESOURCE_PREFIX}-hub --capability-name argocd --query 'capability.configuration.argoCd.serverUrl' --output text 2>/dev/null || echo "")
+
+# If EKS capability ArgoCD URL is available, use it; otherwise fall back to ingress domain
+if [ -n "$ARGOCD_SERVER_URL" ] && [ "$ARGOCD_SERVER_URL" != "None" ]; then
+  ARGOCD_URL="$ARGOCD_SERVER_URL"
+  print_info "Using EKS-managed ArgoCD URL: $ARGOCD_URL"
+else
+  ARGOCD_URL="https://$DOMAIN_NAME/argocd"
```
hmuthusamy, line 61: @allamand  Are we still setting up argocd during initial bootstrap? Do we need to set this condition with capabilities enabled?

gitops/addons/default/addons/argo-cd/values.yaml
```diff
@@ -6,7 +6,7 @@ global:
   nodeSelector:
     eks.amazonaws.com/compute-type: auto
     kubernetes.io/os: linux
-    karpenter.sh/nodepool: system
+    karpenter.sh/nodepool: system-peeks
```
shapirov103, line 9: is it needed after the move to capabilties?

allamand: This is related to auto mode, not EKS Capabilities. We want to rely on our own node pools, and not using the default EKS auto mode ones (so we con configure consolidation windows and other parameters).

We still have a chicken & eggs problem, we need to keep the system nodepool from EKS auto mode to deploy gitlab, which is then used as our gitops source.

shapirov103: Got it, makes sense.

platform/infra/terraform/scripts/1-tools-urls.sh
allamand: yes we still setup the variable with capabilities, but you are right that the fallback on ingress is not necessary and should be removed, i'll take care of it

PR #485, opened 2026-03-02
improve script to not fail on outofSync healthy apps

allamand:
*Issue #, if available:*

*Description of changes:*

With new EKS version, we have false out of sync on backstage on keycloack (kubernetes add null annotation timestamp).
The scripts was blocking on thoses false out of sync issue.

Now it check last success to know if it considere it ok or not


elamaran11 (approved):
LGTM. Is this tested?

PR #486, opened 2026-03-02
feat: integrate Identity Center and ArgoCD token automation into bootstrap flow

punkwalker:
- Add configure_identity_center.py to set up Keycloak as external IdP for IAM Identity Center
- Add argocd_token_automation.py to retrieve ArgoCD auth tokens via Keycloak SSO
- Integrate both scripts into 0-init.sh between ArgoCD readiness and app dependency checks
- Extract wait_for_keycloak_ready() into argocd-utils.sh as reusable function
- Export ARGOCD_AUTH_TOKEN and ARGOCD_SERVER env vars for downstream scripts
- Persist token vars to ~/.bashrc.d/platform.sh via update_workshop_var
- Add 2 new Users and Groups in Keycloak for Editor and Viewer roles and Update IAM IDC group names to match Keycloak group names.
- Update identity-center/main.tf with SSO instance configuration
- configure_identity_center.py runs with 3 retries, 5-min wait, warning-only on failure

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/gitops/addons/charts/keycloak/templates/keycloak-config.yaml
+++ b/gitops/addons/charts/keycloak/templates/keycloak-config.yaml
@@ -143,7 +143,7 @@ data:
         "username": "user3",
         "email": "user3@example.com",
         "firstName": "user",
-        "lastName": "two",
+        "lastName": "three",
         "requiredActions": [],
         "emailVerified": true,
         "groups": [
--- a/platform/infra/terraform/identity-center/deploy.sh
+++ b/platform/infra/terraform/identity-center/deploy.sh
@@ -39,7 +39,7 @@ if [[ -n "$INSTANCE_ARN" && "$INSTANCE_ARN" != "null" ]]; then
   echo "export TF_VAR_identity_center_instance_arn=\"$INSTANCE_ARN\""
   echo "export TF_VAR_identity_center_admin_group_id=\"$ADMIN_GROUP\""
   echo "export TF_VAR_identity_center_developer_group_id=\"$DEV_GROUP\""
-  echo "export TF_VAR_identity_center_non_developer_group_id=\"$DEV_GROUP\""
+  echo "export TF_VAR_identity_center_non_developer_group_id=\"$NON_DEV_GROUP\""
   echo ""
   echo "💡 Copy and run the above export commands before deploying clusters."
 else
--- a/platform/infra/terraform/identity-center/main.tf
+++ b/platform/infra/terraform/identity-center/main.tf
@@ -15,7 +15,7 @@ resource "aws_identitystore_group" "admin" {
   description       = "Platform Admin group for EKS Managed Capability for ArgoCD"
 }
 
-# Create admin group
+# Create developer group
 resource "aws_identitystore_group" "editor" {
   count             = local.identity_store_id != null ? 1 : 0
   identity_store_id = local.identity_store_id
@@ -23,7 +23,7 @@ resource "aws_identitystore_group" "editor" {
   description       = "Developer group for EKS Managed Capability for ArgoCD"
 }
 
-# Create developer group
+# Create non-developer group
 resource "aws_identitystore_group" "viewer" {
   count             = local.identity_store_id != null ? 1 : 0
   identity_store_id = local.identity_store_id


gitops/addons/charts/keycloak/templates/keycloak-config.yaml
```diff
@@ -123,6 +127,27 @@ data:
...
+            "temporary": false
+          }
+        ]
+    }
+  user-user3.json: |
+    {
+        "username": "user3",
+        "email": "user3@example.com",
+        "firstName": "user",
+        "lastName": "two",
```
allamand: Three

gitops/addons/charts/keycloak/templates/keycloak-config.yaml
```diff
@@ -123,6 +127,27 @@ data:
         "requiredActions": [],
         "emailVerified": true,
         "groups": [
+          "/editor"
+        ],
+        "enabled": true,
+        "credentials":  [
+          {
+            "type": "password",
+            "value": "${USER1_PASSWORD}",
```
shapirov103, line 136: is USER1_PASSWORD reused for all users? this is user2 section.

platform/infra/terraform/identity-center/deploy.sh
```diff
@@ -33,11 +33,13 @@ echo "📋 Identity Center Configuration:"
 INSTANCE_ARN=$(terraform output -raw instance_arn 2>/dev/null || echo "")
 ADMIN_GROUP=$(terraform output -raw admin_group_id 2>/dev/null || echo "")
 DEV_GROUP=$(terraform output -raw developer_group_id 2>/dev/null || echo "")
+NON_DEV_GROUP=$(terraform output -raw non_developer_group_id 2>/dev/null || echo "")
```
shapirov103, line 36: why is non dev group needed?

platform/infra/terraform/identity-center/deploy.sh
```diff
@@ -33,11 +33,13 @@ echo "📋 Identity Center Configuration:"
...
 
 if [[ -n "$INSTANCE_ARN" && "$INSTANCE_ARN" != "null" ]]; then
   echo "export TF_VAR_identity_center_instance_arn=\"$INSTANCE_ARN\""
   echo "export TF_VAR_identity_center_admin_group_id=\"$ADMIN_GROUP\""
   echo "export TF_VAR_identity_center_developer_group_id=\"$DEV_GROUP\""
+  echo "export TF_VAR_identity_center_non_developer_group_id=\"$DEV_GROUP\""
```
shapirov103: did you mean to reference NON_DEV_GROUP here? - looks like a copy paste error.

platform/infra/terraform/scripts/0-init.sh
```diff
@@ -153,6 +153,107 @@ main() {
         fi
     fi
 
+    # ---------------------------------------------------------------------------
```
shapirov103, line 163: Looks like an excellent candidate to me moved out of solution to the workshop specific pattern.

gitops/addons/charts/keycloak/templates/keycloak-config.yaml
allamand: I think it use same password for all

platform/infra/terraform/scripts/0-init.sh
allamand: This whole scripts should only needed for workshop

gitops/addons/charts/keycloak/templates/keycloak-config.yaml
shapirov103: the naming is misleading. Either the var should be named USER_PASSWORD (or USER_INIT_PWD), or we should have separate vars for each, even if initialized to the same. 
I happy to defer resolution on this, just pointing things that appear odd to an observer due to the "emerging" nature of the workshop.

shapirov103:
@punkwalker conflicts on this branch >>

platform/infra/terraform/scripts/0-init.sh
shapirov103: > This whole scripts should only needed for workshop

Agree. I also observe that most changes trigger cascade changes in the scripts. Let's think about if there is a better way to handle these other than bash scripting.

platform/infra/terraform/identity-center/main.tf
```diff
@@ -11,19 +11,27 @@ locals {
 resource "aws_identitystore_group" "admin" {
   count             = local.identity_store_id != null ? 1 : 0
   identity_store_id = local.identity_store_id
-  display_name      = "eks-argocd-admins"
-  description       = "Admin group for EKS Managed Capability for ArgoCD"
+  display_name      = "admin"
+  description       = "Platform Admin group for EKS Managed Capability for ArgoCD"
+}
+
+# Create admin group
```
csantanapr: I think this comment is developer group (editor)

platform/infra/terraform/identity-center/main.tf
```diff
@@ -11,19 +11,27 @@ locals {
...
+resource "aws_identitystore_group" "editor" {
+  count             = local.identity_store_id != null ? 1 : 0
+  identity_store_id = local.identity_store_id
+  display_name      = "editor"
+  description       = "Developer group for EKS Managed Capability for ArgoCD"
 }
 
 # Create developer group
```
csantanapr: I think this comment is the non_developer (viewer)

platform/infra/terraform/scripts/0-init.sh
```diff
@@ -153,6 +153,107 @@ main() {
...
+                --url "$argocd_server_url" \
+                --username "user1" \
+                --password "${USER1_PASSWORD}" \
+                --output token 2>/dev/null || echo "")
+
+            if [ -n "$argocd_token" ]; then
+                export ARGOCD_AUTH_TOKEN="$argocd_token"
+                export ARGOCD_SERVER=$(echo "$argocd_server_url" | sed 's|https://||' | sed 's|/.*||')
+                print_status "SUCCESS" "ArgoCD auth token retrieved and exported"
+                update_workshop_var "ARGOCD_AUTH_TOKEN" "$ARGOCD_AUTH_TOKEN"
```
csantanapr, line 251: The token is retrieve from the browser login cookie, how long is this token before it expires?
Does this token works for the argocd cli?

platform/infra/terraform/scripts/argocd_token_automation.py
```diff
@@ -0,0 +1,145 @@
...
+import asyncio
+import json
+import os
+import sys
+
+try:
+    from playwright.async_api import async_playwright
+except ImportError:
+    print("Installing playwright...", file=sys.stderr)
+    os.system("pip install playwright >/dev/null 2>&1 && playwright install chromium >/dev/null 2>&1")
```
csantanapr, line 15: This doesn't wrong because playwright, playwright chromium driver, and chromium was installed from the script (configure_identity_center.py) which runs before this script gets called. 
Maybe add a comment here

gitops/addons/charts/keycloak/templates/keycloak-config.yaml
punkwalker: We can tackle this in next PR, We need to streamline user provisioning any how.

platform/infra/terraform/identity-center/deploy.sh
punkwalker: That's third type of access level for Argo CD, probably execs who do not need editor or admin privileges.

platform/infra/terraform/scripts/0-init.sh
punkwalker: I would love to reduce the scripts 😃

platform/infra/terraform/scripts/argocd_token_automation.py
punkwalker: Yes. We are planning to use these scripts separated from each other. Because we need Argo CD token for Backstage - Argo CD integration as well.

platform/infra/terraform/scripts/0-init.sh
allamand: i'll let you do it, i spent too many time ensuring all use cases :)

PR #525, opened 2026-03-11
Fix: CloudFront VPC Origin retry failures & ArgoCD default AppProject

allamand:
*Issue #, if available:*

*Description of changes:*

deployments intermittently fail or deploy in a broken state due to two issues:

1. CloudFront VPC Origin update failures on Terraform retry

When the GitLab Helm chart times out on first deploy (cold cluster, slow image pulls), Terraform retries hit Error: updating CloudFront VPC Origin because the origin is stuck in a transitional state. This cascades into 
exhausting all retry attempts and failing the Bootstrap CodeBuild.

2. ArgoCD default AppProject not created by EKS Capability

The EKS ArgoCD Capability does not reliably create the default AppProject. Without it, ArgoCD cannot reconcile any Applications. The 0-init.sh script waits 15 minutes for Keycloak, times out, then reports success with 0 apps 
deployed — leaving the workshop non-functional while CloudFormation shows UPDATE_COMPLETE.

### DIFF
--- a/platform/infra/terraform/common/gitlab_infra/main.tf
+++ b/platform/infra/terraform/common/gitlab_infra/main.tf
@@ -119,6 +119,9 @@ resource "aws_cloudfront_vpc_origin" "gitlab" {
     replace_triggered_by = [
       kubernetes_service.gitlab_nlb.id
     ]
+    ignore_changes = [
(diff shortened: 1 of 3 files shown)


csantanapr (approved):
LGTM

PR #569, opened 2026-04-15
fix: use long-lived SA token for Backstage kubernetes auth

allamand:
Backstage reads $file: token references once at startup and caches them in memory. The projected SA token (expirationSeconds: 3607) expires after ~1 hour, causing kube:apply scaffolder actions to fail with 401 Unauthorized.

Fix: create a kubernetes.io/service-account-token Secret (non-expiring) and mount it at /var/run/secrets/backstage/. Update all $file: references in k8s-config, kro plugin, and kubernetesIngestor configs to use the new path.

*Issue #, if available:* Fix backstage that could not talk to kubernetes (needed recreation)

*Description of changes:*
use dedicated token

### DIFF
--- a/gitops/addons/charts/backstage/templates/install.yaml
+++ b/gitops/addons/charts/backstage/templates/install.yaml
@@ -9,6 +9,15 @@ metadata:
   name: backstage
   namespace: backstage
 ---
+apiVersion: v1
+kind: Secret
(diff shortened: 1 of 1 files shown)


elamaran11 (approved):
LGTM

PR #570, opened 2026-04-15
fix: remove long-lived SA token, use in-cluster auth for Backstage

allamand:
Remove backstage-sa-token Secret and remaining explicit serviceAccountToken $file: references. Backstage with authProvider: 'serviceAccount' and no explicit token uses InClusterConfig which re-reads the projected SA token automatically on each API call, handling rotation transparently.

*Issue #, if available:*

*Description of changes:*


punkwalker (approved):

csantanapr:
LGTM, we should not provide a value or file referene, it will expire

PR #579, opened 2026-04-28
fix: remove git fetch that fails on NFS-backed filesystems

allamand:
## Problem

On NFS-backed Workshop Studio instances, `0-init.sh` fails with exit code 128:

```
error: unable to write current sha1 into refs/remotes/github/dependabot/npm_and_yarn/applications/next-js/postcss-8.5.10
fatal: renaming 'refs/remotes/origin/dependabot/npm_and_yarn/applications/next-js/postcss-8.5.10' failed
```

The `git fetch origin` in `0-init.sh` and `2-gitlab-init.sh` fetches all remote refs including dependabot branches with long path names that NFS cannot write.

## Fix

Remove unnecessary `git fetch` calls — the repo is already cloned on the correct tag by bootstrap. Just create a local `main` branch from HEAD with `git checkout -B main`.

## Testing

Observed on Workshop Studio instance i-094b4a75328c2df23 (ws1 account) where 0-init.sh failed with exit code 128.

## Companion MR

- platform-engineering-on-eks: https://gitlab.aws.dev/aws-tfc-containers/containers-hands-on-content/platform-engineering-on-eks/-/merge_requests/new?merge_request%5Bsource_branch%5D=fix%2Fgit-clone-depth-and-error-handling

### DIFF
--- a/platform/infra/terraform/scripts/0-init.sh
+++ b/platform/infra/terraform/scripts/0-init.sh
@@ -335,9 +335,10 @@ APPPROJ
     fi
 
     # ---------------------------------------------------------------------------
-    # Phase: GitLab repos setup (must happen BEFORE ArgoCD sync wait)
-    # The bootstrap ArgoCD app points at the GitLab repo — if it's empty,
-    # ArgoCD will loop with "remote repository is empty" for ~15 minutes.
+    # Phase: GitLab repos setup
+    # The platform repo was already pushed to GitLab by deploy.sh (CodeBuild),
+    # giving ArgoCD a head start while the IDE boots. Here we just set up the
+    # local git remote and create the application repos.
     # ---------------------------------------------------------------------------
     # Get GitLab domain from CloudFront distribution
     if [ -z "$GITLAB_DOMAIN" ]; then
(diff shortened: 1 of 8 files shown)


platform/infra/terraform/scripts/0-init.sh
```diff
@@ -388,6 +388,8 @@ APPPROJ
     GITLAB_URL="https://${GIT_USERNAME}:${USER1_PASSWORD}@${GITLAB_DOMAIN}/${GIT_USERNAME}/${WORKING_REPO}.git"
     
     # Preserve GitHub as 'github' remote if origin points to GitHub
+    # Use --no-tags to avoid fetching all GitHub refs (dependabot branches
```
shapirov103: I see --no-tags in the comment but I can't see it applied. Is it by design?

shapirov103 (commented):
See my comment

platform/infra/terraform/scripts/0-init.sh
allamand: I just learn that gitlab revuse us to git push using the shallow clone (--depth=1- so I need to change this in will reintroduce the --no-tags along with --single-branch

allamand: i've updated it, and tested this fixes the issue

shapirov103 (approved):
LGTM

PR #612, opened 2026-05-04
Fix credential expiry in hub:update task

punkwalker:
## Summary
- Adds `credentials:refresh` before the long-running wait in `hub:update`, preventing credential expiry during the 30-minute hub cluster readiness check
- The `install` flow already handles this via `hub:seed` which starts with `credentials:refresh` — this fix closes the same gap in `hub:update`

## Context
Static AWS credentials stored in the `aws-credentials` Secret are a point-in-time snapshot. The `hub:update` task applies Crossplane claims then waits up to 30 minutes for the hub cluster to become Ready. During this window, Crossplane providers make AWS API calls using the Secret, which can fail if the credentials expire.

## Test plan
- [ ] Run `task hub:update` on an EC2 instance and verify credentials are refreshed before the wait
- [ ] Confirm Crossplane providers pick up the updated secret without pod restarts

### DIFF
--- a/cluster-providers/kind-crossplane/Taskfile.yaml
+++ b/cluster-providers/kind-crossplane/Taskfile.yaml
@@ -536,11 +536,9 @@ tasks:
     vars:
       CLEANUP_KIND: '{{.CLEANUP_KIND | default "false"}}'
     cmds:
-      # Refresh kubeconfig and credentials, restart providers to pick up new creds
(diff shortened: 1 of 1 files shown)


allamand (approved):
LGTM

allamand:
@punkwalker please resolve conflicts

shapirov103:
+1 @punkwalker  let's resolve the conflicts

PR #617, opened 2026-05-07
fix(kro): use omit() instead of orValue("") for tracking-id annotation

allamand:
## Summary

When a kro instance (CICDPipeline, AppmodService, etc.) is created without ArgoCD (e.g., manually via kubectl), the `argocd.argoproj.io/tracking-id` annotation doesn't exist on the parent CR.

Previously, `.orValue("")` would set an empty string annotation on all child resources. With `.omit()`, the annotation is simply not set when absent — which is the correct behavior.

## Changes

Replaced `.orValue("")` with `.omit()` for the optional argocd tracking-id annotation across all RGDs:
- cicd-pipeline.yaml (29 occurrences)
- appmod-service.yaml (19)
- rg-eks.yaml (28)
- ray-service.yaml (17)
- rg-vpc.yaml (13)
- rg-eks-vpc.yaml (2)
- rg-s3-bucket.yaml (1)

## Testing

- Applied updated cicd-pipeline RGD on a live EKS cluster with kro as EKS Capability
- RGD remains Active after the change
- Existing CICDPipeline instances (rust, java) remain ACTIVE and healthy
- Child resources with existing tracking-id annotations are unaffected

Ref: https://github.com/kubernetes-sigs/kro/pull/1273#discussion_r3197153696

### DIFF
--- a/gitops/addons/charts/kro/resource-groups/manifests/cicd-pipeline/cicd-pipeline.yaml
+++ b/gitops/addons/charts/kro/resource-groups/manifests/cicd-pipeline/cicd-pipeline.yaml
@@ -108,7 +108,7 @@ spec:
                blockOwnerDeletion: true
                controller: false
           annotations:
-             argocd.argoproj.io/tracking-id: ${schema.metadata.?annotations["argocd.argoproj.io/tracking-id"].orValue("")}
+             argocd.argoproj.io/tracking-id: ${schema.metadata.?annotations["argocd.argoproj.io/tracking-id"].omit()}
              ecr.services.k8s.aws/force-delete: "true"
        spec:
           name: >-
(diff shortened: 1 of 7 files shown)


elamaran11 (approved):
LGTM

gitops/addons/charts/kro/resource-groups/manifests/cicd-pipeline/cicd-pipeline.yaml
```diff
@@ -108,7 +108,7 @@ spec:
                blockOwnerDeletion: true
                controller: false
           annotations:
-             argocd.argoproj.io/tracking-id: ${schema.metadata.?annotations["argocd.argoproj.io/tracking-id"].orValue("")}
+             argocd.argoproj.io/tracking-id: ${schema.metadata.?annotations["argocd.argoproj.io/tracking-id"].omit()}
```
csantanapr, line 111: The whole/field is gone?

PR #618, opened 2026-05-07
fix: add download timeout to model prestage jobs

allamand:
## Summary
Downloads from HuggingFace can stall silently when no HF_TOKEN is set (common in workshop environments). The prestage jobs hang indefinitely, keeping the ArgoCD ray-operator app in Progressing state.

## Changes
- Wrap `huggingface_hub` download with `timeout` command (30 min default)
- Increase `backoffLimit` from 3 to 5 for more retry attempts
- Reduce `activeDeadlineSeconds` from 7200 to 3600

## Testing
Verified on workshop cluster - stalled jobs now get killed and retried automatically.


elamaran11 (approved):
LGTM

PR #614, opened 2026-05-05
CNI Helper chart version update

allamand:
*Issue #, if available:*

*Description of changes:*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #619, opened 2026-05-07
fix: improve Renovate config to detect addon version bumps in addons.yaml

allamand:
## Summary

Fixes Renovate's custom regex to properly detect Helm chart version updates in `gitops/addons/bootstrap/default/addons.yaml`.

## Changes

- **Fix field order regex** — addons.yaml uses `chartName → chartRepository → defaultVersion` but the regex expected `chartName → defaultVersion → chartRepository`. Added both orderings.
- **Add OCI registry support** — new regex patterns for `ghcr.io` and `public.ecr.aws` chart repositories (used by Kargo, Grafana, etc.)
- **Fix dependabot.yml** — remove deprecated `enable-beta-ecosystems` key and add minimal entry to satisfy schema validation.

## Context

Previously, Renovate never opened PRs for addon version bumps (e.g., cni-metrics-helper 1.19.2 → 1.21.1) because the regex didn't match the field order in addons.yaml. These had to be done manually (see PR #614).


hmuthusamy (approved):
LGTM

PR #624, opened 2026-05-08
Revert "fix(kro): use omit() instead of orValue("")" - omit() not supported

allamand:
## Reverts #617

The `.omit()` CEL function is not supported by the kro version deployed via EKS Capabilities. All RGDs using it remain **Inactive** with:

```
found no matching overload for 'omit' applied to 'optional_type(string).()'
```

This blocks the entire workshop — CICDPipeline, AppmodService, S3Bucket, EksCluster, RayService RGDs all fail to activate.

Reverting back to `.orValue("")` which is confirmed working on the deployed kro version.

## Tested

- Applied RGD with `.omit()` → Inactive
- Applied RGD with `.orValue("")` → Active
- `.empty()` also does NOT work (`undeclared reference to 'empty'`)

### DIFF
--- a/gitops/addons/charts/kro/resource-groups/manifests/appmod-service.yaml
+++ b/gitops/addons/charts/kro/resource-groups/manifests/appmod-service.yaml
@@ -131,7 +131,7 @@ spec:
               blockOwnerDeletion: true
               controller: false
           annotations:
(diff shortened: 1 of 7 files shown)


hmuthusamy (approved):
LGTM

PR #626, opened 2026-05-08
fix: add missing resource types to ray-serve AppProject whitelists

allamand:
## Problem

ArgoCD was not displaying KRO child resources in the application tree view for ray-serve applications. Only the parent `kro.run/RayService` was visible, while all child resources (PVC, PV, Ingress, PDB, ray.io/RayService, ray.io/RayCluster) were hidden.

## Root Cause

The AppProject `namespaceResourceWhitelist` in ray-serve Backstage templates was too restrictive — it only allowed a subset of resource types that KRO creates.

## Fix

Added missing resource types to all 6 ray-serve template files (cpu, gpu, trainium × 2 variants):

- `clusterResourceWhitelist`: `PersistentVolume`
- `namespaceResourceWhitelist`: `PersistentVolumeClaim`, `PodDisruptionBudget`
- `template.yaml` variants also needed: `ray.io/RayService`, `ray.io/RayCluster`, `networking.k8s.io/Ingress`

## Testing

Verified on a live cluster by patching the AppProject — ArgoCD immediately showed the full resource tree with all KRO child resources.


csantanapr (approved):
LGTM

PR #630, opened 2026-05-11
fix: replace crontab with systemd timer for ArgoCD token refresh

allamand:
Fixes the cron job added in #629 — `crontab`/`crond` is not available on the Workshop Studio AMI (Amazon Linux 2023).

Replaces with a systemd timer that:
- Fires 5 min after boot + every 4 hours
- Runs as `ec2-user` with full shell environment
- Logs via `journalctl -u argocd-refresh-token.service`

**Tested on ws027 IDE instance** — token refresh completes in ~15s.


csantanapr (approved):
LGTM

PR #635, opened 2026-05-12
fix: use $HOME instead of ~ in systemd ExecStart

allamand:
Follow-up to #634. Tilde expansion is not guaranteed in systemd ExecStart — use $HOME which is explicitly set in the unit file.


csantanapr (commented):
LGTM

csantanapr (approved):

PR #638, opened 2026-05-12
Fix IAM policy dependency to prevent capability unhealthy state during destroy

NicholasBlaskey:
*Issue #, if available:*

*Description of changes:*

Before this change the dependency graph looked like

```
policy depends on role
capability depends on role
```

On delete we deleted capability and policy in parallel leading the capability to become unhealthy while it was deleted.

Add an explicit dependency to make the graph look like
```
policy depends on role
capability depends on policy
capability depends on role
```


allamand (approved):
LGTM, thanks

shapirov103 (approved):
LGTM

PR #620, opened 2026-05-07
chore(deps): major addon chart version upgrades and Backstage code base removal.

allamand:
## ⚠️ Major Version Upgrades — Requires Testing

These are significant version bumps that may include breaking changes (API removals, renamed values, changed defaults).

| Addon | Current | Latest | Risk |
|-------|---------|--------|------|
| external-secrets | 0.19.2 | 2.4.1 | 🔴 Major (0.x → 2.x) |
| kube-state-metrics | 5.25.1 | 7.3.0 | 🔴 Major |
| aws-efs-csi-driver | 3.0.7 | 4.1.0 | 🟡 Major |
| opentelemetry-operator | 0.71.1 | 0.112.1 | 🟡 Large jump |
| external-dns | 1.14.5 | 1.21.1 | 🟡 Minor but large |
| ingress-nginx | 4.12.2 | 4.15.1 | 🟢 Minor |
| kyverno | 3.3.6 | 3.8.0 | 🟢 Minor |
| kyverno-policies | 3.3.6 | 3.8.0 | 🟢 Minor |
| policy-reporter | 3.1.0 | 3.7.4 | 🟢 Minor |
| prometheus-node-exporter | 4.39.0 | 4.55.0 | 🟢 Minor |
| aws-for-fluent-bit | 0.1.34 | 0.2.0 | 🟢 Minor |

Also remove the local backstage build, to use the CNOE version, cf #667 

## Testing Required
- Deploy to a test account and verify all ArgoCD apps sync successfully
- Check for CRD changes (especially external-secrets, kube-state-metrics)
- Verify values.yaml compatibility with new chart versions

PR #676, opened 2026-05-19
fix: end-to-end workshop deployment (GitOps, ExternalSecret diffs, KubeVela, Backstage SSO)

### LINKED ISSUE #675: Workshop deployment broken: GitOps bootstrap, ExternalSecret diffs, and Backstage Keycloak SSO
## Summary

A full end-to-end validation of the platform-engineering-on-eks workshop revealed several issues that prevent a fresh deployment from working. They cascade from a single root cause but each needs its own fix.

## Symptoms

- Fresh deployment: **no ArgoCD Applications get created at all** — spoke clusters stay empty, Backstage/Keycloak/addons never deploy.
- After GitOps recovery: many apps stay stuck `OutOfSync` even though they reconcile cleanly.
- After GitOps recovery: KubeVela Applications fail to deploy on spoke clusters.
- Backstage shows a 'Sign in as Guest' button instead of Keycloak SSO; the proper OIDC flow never works.
- The validation helper (`backstage-auth.sh`) returns `null` instead of a Backstage identity token.

### 1. GitOps bootstrap circular dependency
The bootstrap ApplicationSet template references `project: platform`, but the `platform` AppProject is itself defined in the content the ApplicationSet would sync. Introduced in commit `fd747803` (`feat(argocd): Add scoped AppProjects for platform and spoke-workloads`). With EKS ArgoCD Capability, the ArgoCD instance is fresh and only has the `default` project, so the bootstrap Application is never created.

### 2. external-secrets v2.x default field not in ignoreDifferences
The external-secrets chart was upgraded **0.19.2 → 2.4.1** in PR #620. The v2.x controller defaults `nullBytePolicy: Ignore` on every ExternalSecret at:
- `.spec.data[].remoteRef.nullBytePolicy`
- `.spec.dataFrom[].extract.nullBytePolicy`

This field didn't exist in v0.19.x, so existing `ignoreDifferences` blocks don't account for it, causing persistent OutOfSync diffs across many apps. Several apps (crossplane-aws, kargo, grafana-dashboards, devlake, fleet-secrets) had no ExternalSecret `ignoreDifferences` at all.

### 3. spoke-workloads AppProject blocks Namespace
`spoke-workloads` has `clusterResourceBlacklist` for Namespace. KubeVela creates Namespaces as managed resources when reconciling Application CRs, so deployments fail with `resource :Namespace is not permitted in project spoke-workloads`.

### 4. Backstage Keycloak OIDC sign-in completely broken
Multiple cascading issues:

- **Backend** uses sign-in resolver `emailMatchingUserEntityProfileEmail`, which is not registered for the `keycloak-oidc` provider in the CNOE backstage-app build (it has a hardcoded resolver in `packages/backend/src/plugins/auth.ts`). The provider gets entirely skipped at startup.
- **Frontend** `packages/app/src/App.tsx` hardcodes `<SignInPage providers={['guest']} auto />`, so even when the backend is fixed, the UI shows a Guest button.
- The proper fix needs to import the existing `keycloakOIDCAuthApiRef` from `apis.ts` — calling `createApiRef()` inline crashes the app with a duplicate registration error.

### 5. backstage-auth.sh validation helper broken
The script targets `env=development` while Backstage is configured for `production`, and uses the refresh endpoint that doesn't deliver tokens in this Backstage version. The OIDC token is actually returned URL-encoded inside the frame HTML body.

## Impact

The workshop is non-functional end-to-end. A fresh deployment cannot be validated because the bootstrap never starts, and even with the bootstrap fixed, KubeVela deployments and Backstage authentication don't work.

## Fix

See PR #676 — consolidates fixes for all five issues.

allamand:
Fixes #675

## Summary

Consolidates fixes discovered while validating the workshop end-to-end. A fresh deployment failed to start, and even after the initial blocker was unblocked, several downstream issues kept the platform from being usable.

### GitOps bootstrap can start
The bootstrap ApplicationSet referenced `project: platform`, which is itself defined in the content it would sync — a chicken-and-egg problem on a fresh ArgoCD (EKS Capability). Reverted to `project: default` so the bootstrap Application creates the `platform` and `spoke-workloads` AppProjects (sync-wave 0) before anything else needs them.

### ExternalSecrets stop reporting false OutOfSync
The external-secrets chart upgrade in #620 (0.19.2 → 2.4.1) introduced a new defaulted field `nullBytePolicy: Ignore` at `.spec.data[].remoteRef` and `.spec.dataFrom[].extract`. Added `jqPathExpressions` covering both paths in every ExternalSecret `ignoreDifferences` block, plus added missing blocks for crossplane-aws, kargo, grafana-dashboards, devlake, and fleet-secrets.

### KubeVela can deploy on spoke clusters
Replaced the `Namespace` blacklist in the `spoke-workloads` AppProject with an explicit whitelist. KubeVela creates Namespaces as managed resources when reconciling Application CRs.

### Backstage Keycloak SSO actually works
- **Backend** (`gitops/addons/charts/backstage/templates/install.yaml`): drop guest provider, set `environment: production`, add `signInPage: keycloak-oidc`, drop config-based sign-in resolvers (the CNOE build has a hardcoded resolver in code that's used instead).
- **Frontend** (`backstage/patches/cnoe-customizations.patch`): replace the hardcoded `<SignInPage providers={['guest']} auto />` with a Keycloak provider object, importing the existing `keycloakOIDCAuthApiRef` from `apis.ts` (calling `createApiRef()` inline crashes the app).
- Removed cluster-scoped `ResourceGraphDefinition` from `kubernetesIngestor` resource list (it's namespaced-only).

### S3 bucket templates work via API, not just UI form
The Backstage scaffolder API does not enforce JSON-schema `default` values — only the UI form does. API callers that omit `namespace` ended up with empty strings, causing ArgoCD `InvalidSpecError` and `catalog:register` 400 errors. Apply Nunjucks `| default(...)` filter explicitly on every `parameters.namespace` reference in both `s3-bucket-ack-kro` and `s3-bucket-ack-kubevela` templates.

### Counter-based metrics actually work in metrics-driven rollouts
Phase 30.5 uses `avg(rocket_http_requests_duration_seconds_sum) > 2`, but `_sum` is a Prometheus Counter — `avg()` returns the cumulative value across all time, which only grows. Once a slow deployment runs, the metric stays above the threshold forever, preventing recovery.

Added an optional `rateInterval` field to `AppmodService` `evaluationCriteria`. When present, the metric is wrapped in `rate(metric[interval])` before applying the aggregation function — works correctly for counters. Applied to both KubeVela (CUE) and kro (CEL) templates.

The corresponding workshop content change (filed separately) updates Phase 30.5 to use `rateInterval: 1m` and an appropriate threshold.

### Validation helper extracts the token correctly
Rewrote `scripts/validation/backstage-auth.sh` to target `env=production` and read the OIDC token from the URL-encoded frame HTML body instead of the refresh endpoint, which doesn't deliver tokens in this Backstage version.

## Testing

End-to-end on a live workshop environment:
- All 61 ArgoCD apps Synced + Healthy
- Hub + 2 spoke clusters fully populated with addons
- Module 10 (Platform exploration + S3 bucket via Backstage scaffolder) ✅
- Module 20 (Rust app dev/prod deploy via KubeVela) ✅
- Module 30 (Progressive delivery: blue→yellow, Java with quality gates, Kargo promotion DEV→PROD, metrics-driven rollback) ✅
- Backstage shows a Keycloak SSO sign-in button; OIDC flow completes
- `backstage-auth.sh` returns a valid identity token
- AnalysisRun query for counter metrics now generates `avg(rate(metric[1m]))` instead of bare `avg(metric)`


elamaran11 (approved):
Dont have any specific concerns because the validation is done with test account. @shapirov103 @hmuthusamy Check from your end.

backstage/patches/cnoe-customizations.patch
shapirov103, line 1: What is this file for? Is it leveraged?

gitops/addons/charts/backstage/templates/install.yaml
```diff
@@ -202,14 +202,13 @@ data:
         type: 'local' # Alternatives - 'googleGcs' or 'awsS3'. Read documentation for using alternatives.
 
     auth:
-      environment: development
+      environment: production
```
shapirov103, line 205: Unclear on this change. I suppose the point was to undo the development mode, is that correct? 

Why not make these the same as our environments? In the context of the solution "production" maybe confused with spoke production. Can we call this one "hub" or does carry some semantic meaning?

gitops/addons/charts/kro/resource-groups/manifests/appmod-service.yaml
```diff
@@ -844,7 +844,7 @@ spec:
...
...) : 1, 'successCondition': 'result[0] ' + c.comparisonType + ' ' + c.threshold, 'provider': {'prometheus': {'address': '{{args.amp-workspace-url}}', 'query': (has(c.function) ? c.function + '(' : '') + (has(c.rateInterval) ? 'rate(' : '') + c.metric + '{container_name=\"' + schema.spec.image_name + '\", namespace=\"' + schema.metadata.namespace + '\"}' + (has(c.rateInterval) ? '[' + c.rateInterval + '])' : '') + (has(c.function) ? ')' : ''), 'authentication': {'sigv4': {'region': '{{args.amp-workspace-region}}'}}}}} : {'name': 'metric-' + c.successOrFailCondition + '-' + c.metric + '-' + schema.metadata.name, 'interval': c.interval, 'count': has(c.count) ? int(c.count) : 1, 'failureCondition': 'result[0] ' + c.comparisonType + ' ' + c.threshold, 'provider': {'prometheus': {'address': '{{args.amp-workspace-url}}', 'query': (has(c.function) ? c.function + '(' : '') + (has(c.rateInterval) ? 'rate(' : '') + c.metric + '{container_name=\"' + schema.spec.image_name + '\", namespace=\"' + schema.metadata.namespace + '\"}' + (has(c.rateInterval) ? '[' + c.rateInterval + '])' : '') + (has(c.function) ? ')' : ''), 'authentication': {'sigv4': {'region': '{{args.amp-workspace-region}}'}}}}})}"
```
shapirov103, line 847: was the a functional change here? did it fail before?

shapirov103 (commented):
Looks good, a couple of questions

backstage/patches/cnoe-customizations.patch
allamand: yes we used it to apply custom backstage configuration from CNOE backstage app. I think this is temporary, as we may see how to apply our needed changes directly into CNOE backstage image

gitops/addons/charts/backstage/templates/install.yaml
allamand: no, this is specific to backstage. In development mode, there is a guest user that we can't remove. passing it to production mode let backstage work for production, and then guest user is deactivated, (and maybe other features also).

gitops/addons/charts/kro/resource-groups/manifests/appmod-service.yaml
allamand: the main change here, we used a rate() to limit the scope of the fail on timeline. without that, it was just counting all failures from the start, and then it was not possible to come back to the normal state

gitops/addons/charts/backstage/templates/install.yaml
shapirov103: From what I understand, as long as the environment is NOT development, the guest user is removed. I don't think the name "production" has a meaning, seems like this field is just a bespoke string.

allamand: oh that is possible. but is it an issue to have production ? users won't see thoses anyway

shapirov103: Not an issue at the moment, just caused of confusion since we have a spoke for production environment and this is not it. We can revisit if it causes confusion for anyone else. The terminology overlaps.

shapirov103 (approved):
LGTM

PR #678, opened 2026-05-20
fix(observability): guard spoke scrapers against missing/None securityGroupIds

allamand:
Prevents AMP scrapers from being created with invalid security group IDs when spoke cluster annotations are not yet populated.

The template now skips scraper generation when securityGroupIds is empty or None. ArgoCD auto-sync will create the scrapers once valid values are available via the ApplicationSet annotation re-render.


elamaran11 (approved):
LGTM

hmuthusamy (approved):
LGTM

PR #706, opened 2026-05-27
fix(pod-identities): Guard PodIdentityAssociation against empty region

hmuthusamy:
The EKS ArgoCD Capability does not reliably merge valuesObject over chart defaults for scalar values. When aws.region defaults to empty string in values.yaml, the rendered PodIdentityAssociation has no region field, failing CRD validation and blocking the entire pod-identities app from syncing (including LBC and external-dns).

Changes:
- Remove explicit empty-string defaults for aws (region, clusterName, accountId) from chart values.yaml — use aws: {} instead so valuesObject is the sole source
- Add aws.region guard to the PodIdentityAssociation template conditional — if region is empty, skip rendering instead of producing an invalid manifest

*Issue #, if available:*

*Description of changes:*


shapirov103 (approved):
LGTM

PR #709, opened 2026-05-27
feat(cloudfront): CloudFront exposure mode for domain-less deployments

allamand:
## CloudFront exposure mode for domain-less deployments

Enables the platform to run without a custom domain by using CloudFront distributions for HTTPS ingress.

### Key Changes

**Infrastructure & Bootstrap**
- Crossplane NAT Gateway fix: composite field patching to avoid LateInitialize race condition
- Taskfile: POSIX-compatible `create-mgmt-roles`, fix elapsed time display, add `spokes:enable-crossplane`/`spokes:enable-kro` tasks
- Add `secrets-manager:seed-secrets` task for platform user credentials
- Add `sts:AssumeRole` policy to ACK capability role for cluster-mgmt roles
- Add `eks:*` permissions to `cluster-mgmt-eks` role

**ArgoCD 3.x (EKS Capability) Compatibility**
- Replace `dig` with `index`/`or` in ApplicationSet templates (ArgoCD 3.x strict typing)
- Add `project: default` to cluster secrets (required by ArgoCD 3.x)
- Use `index` for optional annotations with `missingkey=error`
- Fix GitLab Keycloak client `redirectUri` to use GitLab CloudFront domain

**Fleet Management & Spoke Clusters**
- Rename `kro-values/` → `spoke-values/`, separate `crossplane-clusters/` and `kro-clusters/`
- `clusters-kro` ApplicationSet: one app per cluster (aligned with main), `CreateNamespace=true`
- Add overlay source (GitLab fleet-config) to cluster ApplicationSets
- Fleet README documenting Crossplane vs KRO provisioning
- Add `secretsmanager` IAMRoleSelector for ACK spoke provisioning
- Fix `resourcePrefix` vs `clusterName` mismatch in IAMRoleSelectors
- Align `rg-eks.yaml` and `rg-eks-vpc.yaml` with main (remove `enable_ack_*`, single NAT Gateway, camelCase annotations)

**GitLab**
- Fix `ROOT_GITLAB_TOKEN` usage in init job (was using raw password as API token)
- Add `aws-load-balancer-name` annotation for NLB adoption
- Fix `gitlab_domain_name` in Keycloak client redirect URI

**Observability**
- Increase cert-manager-cainjector memory to 512Mi (OOMKilled at 128Mi)
- Add `ttlStrategy` to devlake workflow for pod cleanup
- Enable observability defaults (cni_metrics, prometheus_node_exporter, kube_state_metrics)
- Fix `readyWhen`: `eks.status.clusterState` instead of `eks.status.state`

**Documentation**
- Steering: ACK, ArgoCD 3.x, Crossplane, KRO troubleshooting guidance
- Fleet README with cluster provisioning instructions

---

### Additional fixes (follow-up commits)

**Bootstrap robustness**
- Prevent Crossplane CRD-ordering race and AWS LB Controller webhook cert drift: PreSync CRD-readiness gate on `crossplane-pod-identity`, `enableCertManager` for the LBC webhook, `pod_identities` moved after `crossplane-base` in sync-wave, and `wait-for-full-sync` now surfaces per-app errors and fails fast on CRD-ordering failures.
- Correct `EksclusterWithVpc` RGD default `addonsRepoBasePath` (`gitops/addons/` → `gitops/`); the doubled value produced non-existent overlay paths (`gitops/addons/addons/...`) that left all of a spoke's apps `Unknown`.

**Spoke provisioning (kro / Crossplane)**
- kro spokes inherit the control-plane's deploy revision (`addonsRepoRevision`/`fleetRepoRevision`) via `clusters-kro` global values instead of defaulting to `main` (charts that exist only on the deploy branch were unresolvable on spokes).
- `clusters-kro` git generator reads per-cluster trigger JSONs from the overlay (`fleet-config`) with a base fallback, so `task spokes:enable-kro` actually drives generation; removed the redundant base-repo `kro-clusters/clusters/peeks-spoke-prod.json`.
- Label fleet-secret (direct-mode / Crossplane) spoke cluster secrets with `fleet_member=spoke`, matching kro-provisioned spokes.

**Developer experience / docs**
- `spokes:enable-crossplane` / `spokes:enable-kro` refresh the local `~/environment/fleet-config` checkout after pushing (best-effort, non-fatal).
- fleet-config overlay README is now seeded from the platform source (`platform/fleet-config/README.md`) by the GitLab init-job (single source of truth), replacing the inline placeholder.

**Spoke access entries & Crossplane provider bootstrap**
- Emit `hub_cluster_name` on the KRO `EksCluster` `argocdSecret` so the appset-chart `destinationHub` resolution targets the hub. The `pod-identities` bootstrap app now deploys on the hub (whose Crossplane providers already have credentials) and creates the spoke's `provider-aws-iam`/`provider-aws-eks` Pod Identity associations, instead of falling back to the credential-less spoke. This chicken-and-egg previously left `crossplane-base`/`pod-identities` `Degraded` on KRO-provisioned spokes (`no EC2 IMDS role found`). Dev worked only because its Crossplane `fleet-secret` already carried the annotation.
- Add a `grant-ide-access` task (and extend `kubeconfig`) to grant the current caller — the IDE EC2 instance role (`*SharedRole*`) — cluster-admin access entries on the hub and every `<prefix>-*` cluster. `create-config.sh` deliberately excludes the instance role from `adminRoleName`, so it had no access entry on KRO-provisioned spokes (`kubectl` 401). Idempotent and additive.
- Evolve `adminRoleName` (string) → `adminRoles` (`[]{label, roleName}`) on the `EksCluster`/`EksclusterWithVpc` RGDs, turning the `accessEntryAdmin` resource into a KRO v0.9 `forEach` that creates one cluster-admin `AccessEntry` per role — so a cluster can grant the console/participant role and the IDE/instance role at the same time. `label` is a DNS-1123-safe discriminator (IAM role names may contain uppercase/illegal chars). Backward-compatible: `clusters.yaml` wraps a legacy single `adminRoleName` into a one-element list.


config.yaml
```diff
@@ -49,3 +49,14 @@ identityCenter:
 # EKS ArgoCD Capability
 argocdCapability:
   name: "argocd"
+
+# Exposure mode: "domain" (HTTPS + ACM cert + custom domain) or "cloudfront" (HTTP ALB behind CloudFront)
+exposure:
+  mode: "domain"
```
punkwalker: Instead of adding new config field, can we use .domain=<user-domain>/cloudfront? When it's cloudFront we use same decision making in templates to create cloudFront?

P.S. If we are pre-creating the CloudFront domain before we run task install, then we can configure templates to drop hosts parameter of ingresses if the domain name contains *$cloudfront.net". Can something like this be done?
Ref:
```
      INGRESS_DOMAIN:
        sh: |
          if [ "{{.EXPOSURE_MODE}}" = "cloudfront" ] && [ -f {{.ROOT_DIR}}/private/cloudfront-domain ]; then
            cat {{.ROOT_DIR}}/private/cloudfront-domain
          else
            echo "{{.DOMAIN}}"
          fi
      ADMIN_ROLE_NAME:
        sh: yq '.adminRoleName // ""' {{.CONFIG_FILE}}
      GITLAB_DOMAIN:
        sh: |
          if [ "{{.EXPOSURE_MODE}}" = "cloudfront" ] && [ -f {{.ROOT_DIR}}/private/gitlab-cloudfront-domain ]; then
            cat {{.ROOT_DIR}}/private/gitlab-cloudfront-domain
          fi
```

config.yaml
```diff
@@ -49,3 +49,14 @@ identityCenter:
...
+
+# Console admin role name (used by Backstage templates and added as cluster admin if specified)
+adminRoleName: "<console-role-name>"       # e.g. WSParticipantRole
+
+# ML model S3 bucket (created by Taskfile if enabled)
+modelS3Bucket:
+  enabled: false
```
punkwalker, line 74: Is this something that can be created by the Application Abstractions? instead of creating from local commands?

config.yaml
```diff
@@ -49,3 +49,14 @@ identityCenter:
 # EKS ArgoCD Capability
 argocdCapability:
   name: "argocd"
+
+# Exposure mode: "domain" (HTTPS + ACM cert + custom domain) or "cloudfront" (HTTP ALB behind CloudFront)
+exposure:
+  mode: "domain"
+
+# Console admin role name (used by Backstage templates and added as cluster admin if specified)
+adminRoleName: "<console-role-name>"       # e.g. WSParticipantRole
```
punkwalker: Can we not deduce the roleName from aws sts get-caller-identity? Maybe I am missing some context here

cluster-providers/kind-crossplane/Taskfile.yaml
```diff
@@ -365,6 +377,13 @@ tasks:
       - kubectl -n crossplane-system wait --for=condition=Ready --timeout=300s roles.iam.aws.upbound.io --all
       - kubectl -n crossplane-system wait --for=condition=Ready --timeout=600s podidentityassociations.eks.aws.upbound.io --all
       - task: argocd:capability
+      - cmd: |
+          if [ "{{.EXPOSURE_MODE}}" = "cloudfront" ]; then
+            cd {{.ROOT_DIR}} && task kind-crossplane:hub:create-alb
+            cd {{.ROOT_DIR}} && task kind-crossplane:hub:cloudfront
+            cd {{.ROOT_DIR}} && task kind-crossplane:gitlab:create-nlb
+            cd {{.ROOT_DIR}} && task kind-crossplane:gitlab:cloudfront
+          fi
```
punkwalker: If these commands are ran together always, then it's better to it this way by grouping them in a separate task:
```
version: '3'

tasks:
  # Your primary entry point task
  setup-exposure:
    cmds:
      - task: deploy-cloudfront-resources

  # The conditional subtask running your local commands
  deploy-cloudfront-resources:
    if: '[ "{{.EXPOSURE_MODE}}" = "cloudfront" ]'
    cmds:
      - task: hub:create-alb
      - task: hub:cloudfront
      - task: gitlab:create-nlb
      - task: gitlab:cloudfront
```

We can have native task command condition instead of bash conditions

cluster-providers/kind-crossplane/Taskfile.yaml
```diff
@@ -351,6 +355,14 @@ tasks:
...
           aws eks associate-access-policy --cluster-name {{.HUB_CLUSTER_NAME}} --principal-arn "$PRINCIPAL_ARN" --policy-arn arn:aws:eks::aws:cluster-access-policy/AmazonEKSClusterAdminPolicy --access-scope type=cluster --region {{.AWS_REGION}} --output text 2>/dev/null || true
           echo "Access entry created for $PRINCIPAL_ARN"
+      - cmd: |
+          ADMIN_ROLE=$(yq '.adminRoleName // ""' {{.CONFIG_FILE}})
+          if [ -n "$ADMIN_ROLE" ]; then
+            ADMIN_ARN="arn:aws:iam::{{.AWS_ACCOUNT_ID}}:role/$ADMIN_ROLE"
+            aws eks create-access-entry --cluster-name {{.HUB_CLUSTER_NAME}} --principal-arn "$ADMIN_ARN" --type STANDARD --region {{.AWS_REGION}} --output text 2>/dev/null || true
+            aws eks associate-access-policy --cluster-name {{.HUB_CLUSTER_NAME}} --principal-arn "$ADMIN_ARN" --policy-arn arn:aws:eks::aws:cluster-access-policy/AmazonEKSClusterAdminPolicy --access-scope type=cluster --region {{.AWS_REGION}} --output text 2>/dev/null || true
+            echo "Admin access entry created for $ADMIN_ARN"
+          fi
```
punkwalker, line 482: I think, we create the Access Entry for the IAM entity running the task commands, if we want, probably we should call it additionalAdminRole. Also, if this is specific to the workshop, can we not create this on workshop bootstrap side?

config.yaml
allamand: this needs to be done pre-gitops, so make sense to have this in the taskfile. This is used when user want to pre-download some models in s3 for faster starts.

config.yaml
allamand: no we can't, RoleName used in AWS console, and in the IDE are generally different

cluster-providers/kind-crossplane/Taskfile.yaml
allamand: Not sure it is specific for workshop, but needed if IAM role used in AWS console and IDE are different. We can't do this in bootstrap, as cluster don't exists yet

cluster-providers/kind-crossplane/Taskfile.yaml
allamand: good idea i'll look at it

config.yaml
allamand: maybe, let me look at this

cluster-providers/kind-crossplane/Taskfile.yaml
punkwalker: Then we can probbaly add this into PlatformCluster Composition as additional admin roles which creates Access Entry, I think this can be useful for other use-cases of spokes

allamand: yes and it is already the case for spoke clusters, make sense to have it same way for hub

allamand:
I've made good progress on this: cluster installation and all addons deployment OK.

Now needs to configure backstage to read from github, an from configmap for the dynamic configuration part that was initially done in gitlab custom

allamand:
Need to fix Gitlab SSO

cluster-providers/kind-crossplane/Taskfile.yaml
```diff
@@ -898,6 +1017,296 @@ tasks:
...
+        )
+        CF_CONFIG=$(echo "$CF_CONFIG" | sed "s|ALB_DNS_PLACEHOLDER|$ALB_DNS|g; s|TIMESTAMP|$(date +%s)|g")
+        CF_DOMAIN=$(aws cloudfront create-distribution --distribution-config "$CF_CONFIG" --query "Distribution.DomainName" --output text)
+        mkdir -p {{.ROOT_DIR}}/private
+        echo -n "$CF_DOMAIN" > {{.ROOT_DIR}}/private/cloudfront-domain
+        echo "CloudFront: $CF_DOMAIN"
+      - printf '{{.C_OK}}✓ Platform CloudFront ready.{{.C_RESET}}\n'
+
+  gitlab:create-nlb:
+    desc: Pre-create GitLab NLB for CloudFront (cloudfront mode only)
```
punkwalker: IMO, we should keep this outside of github taskfile, as this is only for the workshop

gitops/addons/charts/keycloak/templates/keycloak-config.yaml
```diff
@@ -963,134 +963,36 @@ spec:
...
-          property: LANGFUSE_CLIENT_SECRET
-    - conversionStrategy: None
-      match:
-        secretKey: LANGFUSE_CLIENT_ID
-        remoteRef:
-          remoteKey: {{ .Values.aws.cluster_name}}/keycloak-clients
-          property: LANGFUSE_CLIENT_ID
+              echo "✓ Kubernetes secret created"
+
+              # Push to AWS Secrets Manager (synchronous — no PushSecret race condition)
```
punkwalker, line 953: May be I am missing something, but why are we not using PushSecret instead of using aws cli?

gitops/addons/configs/backstage/values.yaml
```diff
@@ -1,7 +1,7 @@
 # Backstage configuration
 # Override these in overlays/environments/<env>/backstage/values.yaml
 
-backstage_image: "ghcr.io/backstage/backstage:latest"
+backstage_image: "public.ecr.aws/seb-demo/backstage:cnoe"
 git_username: ""        # Git username for repo access
 working_repo: ""        # Working repository URL
-gitlab_domain_name: ""  # GitLab domain (defaults to ingress_domain_name if same host)
+gitlab_domain_name: "dfzsl5lc0kh4k.cloudfront.net"  # GitLab domain (defaults to ingress_domain_name if same host)
```
punkwalker: Should these values be here? Looks like test env values

gitops/overlays/environments/control-plane/gitlab/values.yaml
```diff
@@ -0,0 +1 @@
+initial_root_password: "TchydtrLg27QJ03i"
```
punkwalker: Should be removed?

platform/infra/terraform/common/ingress-nginx.tf
```diff
@@ -97,7 +97,7 @@ resource "helm_release" "ingress_nginx" {
   name       = "ingress-nginx"
   repository = "https://kubernetes.github.io/ingress-nginx"
   chart      = "ingress-nginx"
-  version    = "4.12.2"
+  version    = "4.15.1"
   namespace  = kubernetes_namespace.ingress_nginx.metadata[0].name
```
punkwalker: Are we using this in this version?

config.schema.json
```diff
@@ -160,6 +160,17 @@
...
+      "type": "object",
+      "properties": {
+        "mode": {
+          "type": "string",
+          "enum": ["domain", "cloudfront"],
+          "default": "domain",
+          "description": "Exposure mode. 'domain' uses HTTPS with ACM cert and custom domain. 'cloudfront' uses HTTP ALB behind CloudFront (no custom domain needed)."
+        }
+      }
+    },
```
punkwalker: Looks like the .exposure field is now removed, right?

punkwalker (commented):
PR is Great! But I feel some of the workshop specific task steps and scripts are creeping into this. @allamand Could you please evaluate if the workshop specific configuration tasks/steps/scripts like Gitlab, ~/.bashrc.d/platform.sh creation etc, can be moved out of this PR?

config.schema.json
allamand: Good catch — the `.exposure` field was vestigial. Exposure mode is now derived automatically from `.domain` (empty `domain` → CloudFront mode, otherwise domain mode) via the `EXPOSURE_MODE` var in the Taskfile. Removed the unused `exposure` block from the schema.

gitops/overlays/environments/control-plane/gitlab/values.yaml
allamand: Yes — removed. The hardcoded `initial_root_password` is gone; the GitLab root password is now sourced from `git_token` (= the workshop USER_PASSWORD / IDE_PASSWORD) via the gitlab registry valuesObject. `gitlab.yaml` and `init-job.yaml` now use `git_token` (with `initial_root_password` kept only as a fallback).

cluster-providers/kind-crossplane/Taskfile.yaml
allamand: Done — the ALB/CloudFront/NLB setup is now grouped in the `hub:setup-cloudfront-exposure` task, which uses a native `status:` condition (`[ "{{.EXPOSURE_MODE}}" != "cloudfront" ]`) to run only in CloudFront mode, instead of scattered inline bash conditions.

config.yaml
allamand: Adopted — `EXPOSURE_MODE` is now derived from `.domain`: leaving `domain` empty selects CloudFront mode (domain auto-created), otherwise domain mode. The separate exposure field has been removed.

gitops/addons/configs/backstage/values.yaml
allamand: Right — that was a leftover test value. The base default is now empty (`gitlab_domain_name: ""`); the real value is injected per-deployment via the registry valuesObject from cluster metadata (`metadata.annotations.gitlab_domain_name`), falling back to `ingress_domain_name`.

platform/infra/terraform/common/ingress-nginx.tf
allamand: No — this `terraform/common` path (ingress-nginx + the CloudFront origin pointing at its NLB) is the legacy Terraform exposure path. In this branch the platform is exposed via ALB (the kind-crossplane Taskfile pre-creates the ALB + CloudFront; the GitOps ingresses use the ALB ingress class). The nginx module isn't used in this flow — we'll clean it up in a separate step.

gitops/addons/charts/keycloak/templates/keycloak-config.yaml
allamand: This is intentional: the OIDC client secrets must be in AWS Secrets Manager synchronously at bootstrap so downstream apps (backstage, jupyterhub, etc.) can consume them on first sync. PushSecret is reconciled asynchronously by ESO, which caused a race condition at first boot. The `aws` CLI push is synchronous and avoids that (noted in the code comment). Open to switching to PushSecret if we add a readiness wait on the consumer side.

platform/infra/terraform/scripts/configure_identity_center.py
```diff
@@ -79,17 +79,45 @@
...
+        fresh = json.dumps({
+            "AccessKeyId": creds_json["AccessKeyId"],
+            "SecretAccessKey": creds_json["SecretAccessKey"],
+            "SessionToken": creds_json["Token"],
+            "Expiration": creds_json["Expiration"],
+        })
+        param_name = f"/{prefix}/keycloak-idc-integration-credentials"
+        ssm.put_parameter(Name=param_name, Value=fresh, Type="SecureString", Overwrite=True)
+        with open(ASSUME_ROLE_CREDENTIALS_FILE, "w") as f:
+            f.write(fresh)
```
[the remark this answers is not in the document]
allamand: These are short-lived IMDS credentials used only by this bootstrap helper. They're stored encrypted in SSM as a `SecureString`, and the local cache file is written with `0o600` (owner-only) under `/tmp` (non-persistent), then consumed as a `credential_process`. Given the ephemeral nature and the encryption/permission controls, we're treating this as accepted risk for the workshop bootstrap and dismissing the CodeQL alert.

cluster-providers/kind-crossplane/Taskfile.yaml
allamand: Done — extracted the CloudFront exposure tasks into a dedicated `Taskfile.cloudfront.yaml`, included under the `cloudfront:` namespace (vars passed via the include). They still run only in CloudFront mode (per-task `status:` guards), and remain Taskfile-driven because the ALB/NLB/CloudFront resources must be pre-created *before* the GitOps bootstrap (the LBC doesn't exist yet). Named by function (`cloudfront`) rather than "workshop", since this is the CloudFront exposure mode and is usable beyond the workshop. Validated with `task --list-all`.

applications/rust/deployment/templates/kro/application.yaml
```diff
@@ -10,7 +10,7 @@ spec:
   port: 8080
   targetPort: 8080
   serviceAccount: rust-service-account-<environment>
-  clusterName: peeks-spoke-dev
+  clusterName: <cluster_name>
```
punkwalker, line 13: Should this be `${CLUSTER_NAME}`, like other apps?

cluster-providers/kind-crossplane/manifests/argocd/create-capability.yaml
```diff
@@ -97,6 +97,61 @@ spec:
                 --policy-arn arn:aws:eks::aws:cluster-access-policy/AmazonEKSClusterAdminPolicy \
                 --access-scope type=cluster 2>/dev/null || true
 
+              echo "=== Create KRO capability ==="
+              KRO_STATUS=$(aws eks describe-capability \
+                --cluster-name "$CLUSTER_NAME" \
+                --capability-name "kro" \
+                --region "$AWS_REGION" \
+                --query 'capability.status' \
```
punkwalker: Looks like we have support for EKS Capability in Crossplane now, Can we move this to XRD?
Ref: https://marketplace.upbound.io/providers/upbound/provider-aws-eks/v2.6.1/resources/eks.aws.m.upbound.io/Capability/v1beta1

cluster-providers/kind-crossplane/Taskfile.cloudfront.yaml
```diff
@@ -0,0 +1,163 @@
...
+
+  setup-exposure:
+    desc: Pre-create ALB and CloudFront distributions (cloudfront mode only)
+    status:
+      - '[ "{{.EXPOSURE_MODE}}" != "cloudfront" ]'
+    cmds:
+      - task: create-alb
+      - task: hub-distribution
+      - task: gitlab-nlb
+      - task: gitlab-distribution
```
punkwalker: We are planning make it hard pre-requisite to create following if user wants to use CloudFront so that we can avoid complicating our solution:

- Create VPC with min 2 private and 2 public subnets
- Create private ALB in the VPC with port 80 listener which returns fixed response (so they don't have to create any TargetGroup). It should have the required tags
```
elbv2.k8s.aws/cluster: <cluster-name>
ingress.k8s.aws/resource: LoadBalancer
ingress.k8s.aws/stack: <Ingress Group Name> (Currently it is HardCoded to "platform")
```
- Create CloudFront VPC origin with ALB as source
- Create CloudFront Distribution and add the VPC origin

After this point, they just give us CloudFrontDomain and VPCId. We take care of bringing everything up.

The command in this taskfile can be a helper script for creating these pre-requisites but they can use any other approach to fulfill the pre-reqs. I know this sounds too much work but it will simplify the installation flow. No Chicken and Egg dependencies.

This will require for us to make following changes to enable use of CloudFront domain:
- The ingresses should have helm if/else to not use ACM certs if it is CloudFront Domain
- The VPCId passed by the user will be used to fetch subnet IDs and these details will be used with separate EKS XRD, which accepts pre-created VPC. 
- The IngressClassParam should switch the scheme to internal when cloudFront domain is used.

cluster-providers/kind-crossplane/Taskfile.cloudfront.yaml
```diff
@@ -0,0 +1,163 @@
...
+# Included by the main Taskfile under the "cloudfront:" namespace.
+tasks:
+  sync-origins:
+    desc: Ensure CloudFront distributions point to the current ALB/NLB DNS (idempotent safety net)
+    cmds:
+      - cmd: |
+          if [ "{{.EXPOSURE_MODE}}" != "cloudfront" ]; then exit 0; fi
+          printf '{{.C_STEP}}▸ Syncing CloudFront origins with current LB DNS...{{.C_RESET}}\n'
+          # Platform ALB
+          ALB_DNS=$(aws elbv2 describe-load-balancers --names "{{.HUB_CLUSTER_NAME}}-platform" --region {{.AWS_REGION}} --query 'LoadBalancers[0].DNSName' --output text 2>/dev/null)
```
punkwalker: As discussed, let's move this to workshop

cluster-providers/kind-crossplane/Taskfile.yaml
```diff
@@ -243,6 +364,8 @@ tasks:
           | kubectl apply --server-side --force-conflicts -f - 2>&1 | grep -v "ProviderConfig" || true
       - kubectl wait --for=condition=Healthy --timeout=300s providers.pkg.crossplane.io --all
       - kubectl wait --for=condition=Healthy --timeout=300s functions.pkg.crossplane.io --all
+      # Grant cluster-admin to provider-kubernetes (needed to manage Jobs, Secrets, etc.)
+      - kubectl create clusterrolebinding provider-kubernetes-admin --clusterrole=cluster-admin --serviceaccount=crossplane-system:provider-kubernetes 2>/dev/null || true
```
punkwalker, line 352: Crossplane Manages k8s Jobs? Are we using kubernetes provider of crossplane?

cluster-providers/kind-crossplane/Taskfile.yaml
```diff
@@ -637,18 +1844,100 @@ tasks:
...
+      KC_DOMAIN:
+        sh: |
+          if [ "{{.EXPOSURE_MODE}}" = "cloudfront" ] && [ -f {{.ROOT_DIR}}/private/cloudfront-domain ]; then
+            cat {{.ROOT_DIR}}/private/cloudfront-domain
+          else
+            echo "{{.DOMAIN}}"
+          fi
     cmds:
       - |
         set -e
```
punkwalker, line 1105: Can't we do this inside the cluster through a Job?

cluster-providers/kind-crossplane/Taskfile.yaml
```diff
@@ -2,6 +2,21 @@ version: "3"
 set: [errexit, nounset, pipefail]
 silent: true
 
+includes:
+  cloudfront:
+    taskfile: ./Taskfile.cloudfront.yaml
+    vars:
+      AWS_REGION: '{{.AWS_REGION}}'
+      HUB_CLUSTER_NAME: '{{.HUB_CLUSTER_NAME}}'
+      ROOT_DIR: '{{.ROOT_DIR}}'
```
punkwalker, line 13: We will keep CloudFront script separate, so we might not need this?

cluster-providers/kind-crossplane/Taskfile.yaml
```diff
@@ -92,20 +109,123 @@ tasks:
...
+        sh: date +%s
     cmds:
       - task: init
+      - 'printf "{{.C_INFO}}  [%ds] init done{{.C_RESET}}\n" "$(($(date +%s) - {{.INSTALL_START}}))"'
       - task: kind:create
+      - 'printf "{{.C_INFO}}  [%ds] kind:create done{{.C_RESET}}\n" "$(($(date +%s) - {{.INSTALL_START}}))"'
       - task: credentials:setup
+      - 'printf "{{.C_INFO}}  [%ds] credentials:setup done{{.C_RESET}}\n" "$(($(date +%s) - {{.INSTALL_START}}))"'
       - task: argocd:install
+      - 'printf "{{.C_INFO}}  [%ds] argocd:install done{{.C_RESET}}\n" "$(($(date +%s) - {{.INSTALL_START}}))"'
```
punkwalker: Let's remove argocd from the kind cluster, as we are not using it.

cluster-providers/kind-crossplane/Taskfile.yaml
```diff
@@ -365,13 +502,82 @@ tasks:
...
       - task: hub:install-eso
       - task: hub:cluster-secret-store
       - task: hub:seed-secret
       - task: hub:apply-root-appset
 
+  hub:create-mgmt-roles:
+    desc: Create cluster-mgmt IAM roles assumed by ACK controllers (single-account prereq)
+    vars:
+      ACK_CAPABILITY_ROLE:
+        sh: aws eks describe-capability --cluster-name {{.HUB_CLUSTER_NAME}} --capability-name ack --region {{.AWS_REGION}} --query 'capability.roleArn' --output text 2>/dev/null | sed 's|.*/||' || echo "{{.HUB_CLUSTER_NAME}}-ArgoCDCapabilityRole"
```
punkwalker, line 516: We should create these management roles from the composition, based on your opinion, either add it as part of composition or helm-chart.

cluster-providers/kind-crossplane/Taskfile.yaml
```diff
@@ -507,10 +714,977 @@ tasks:
...
+  urls:
+    desc: Export platform URLs to ~/.bashrc.d/platform.sh
+    vars:
+      INGRESS_DOMAIN:
+        sh: |
+          if [ "{{.EXPOSURE_MODE}}" = "cloudfront" ] && [ -f {{.ROOT_DIR}}/private/cloudfront-domain ]; then
+            cat {{.ROOT_DIR}}/private/cloudfront-domain
+          else
+            echo "{{.DOMAIN}}"
+          fi
```
punkwalker: We should not modify user files, especially rc files. Let's document how user can access the URLs or user can run this task command to get the urls

cluster-providers/kind-crossplane/Taskfile.yaml
```diff
@@ -507,10 +714,977 @@ tasks:
...
+          printf '{{.C_INFO}}  Skipped marker/CARM cleanup (prune not confirmed). Re-run this task after a successful prune.{{.C_RESET}}\n'
+        fi
+        rm -rf "$REPO_DIR"
+
+  spokes:disable-all:
+    desc: >-
+      DELETE every spoke cluster (KRO + Crossplane) declared in the fleet-config
+      overlay, then wait for them to disappear from AWS. Used as Phase 0 of
+      `destroy` so spokes are removed while the hub controllers (KRO/ACK/Crossplane)
+      are still alive to reconcile their deletion. Idempotent.
```
punkwalker: Why do we have to disable these addons to remove spoke clusters? Can we create a comprehensive destroy script which uses certain sequence of deleting the addon from clusters so that we handle, graceful deletion using Argo CD?

cluster-providers/kind-crossplane/Taskfile.yaml
```diff
@@ -507,10 +714,977 @@ tasks:
...
+          printf '{{.C_ERR}}⚠ Spoke clusters still present after wait:%s{{.C_RESET}}\n' "$STILL_LIVE"
+          printf '{{.C_ERR}}  Proceeding with hub teardown will ORPHAN them. See troubleshooting.md →{{.C_RESET}}\n'
+          printf '{{.C_ERR}}  "Spoke Cluster Stuck Deletion / Recovery". Recommend aborting and fixing first.{{.C_RESET}}\n'
+        fi
+
   hub:restart-identity-pods:
     desc: Restart pods that depend on Pod Identity (external-dns, LBC, Crossplane providers) after identities are ready
     cmds:
       - task: hub:kubeconfig
+      - printf '{{.C_STEP}}▸ Terminating any stuck ArgoCD sync operations on pod-identities apps...{{.C_RESET}}\n'
```
punkwalker, line 834: We should use hook for these restart of the pods. I think,  we have similar mechanism here - https://github.com/aws-samples/appmod-blueprints/tree/feature/agent-platform/gitops/addons/charts/pod-identity-restart-hook

cluster-providers/kind-crossplane/Taskfile.yaml
```diff
@@ -898,6 +2261,106 @@ tasks:
...
       - kind delete cluster --name {{.KIND_CLUSTER_NAME}}
 
+
+
+  idc:configure:
+    desc: Configure IDC with Keycloak as external IdP and sync users via SCIM
+    status:
+      - |
+        IDENTITY_STORE_ID=$(aws sso-admin list-instances --region {{.AWS_REGION}} \
+          --query "Instances[?InstanceArn=='{{.IDC_INSTANCE_ARN}}'].IdentityStoreId" --output text 2>/dev/null)
```
punkwalker: Let's not do the keycloak - IAM IDC integration here and keep it workshop specific. We might end up modifying the user provided IAM IDC.

cluster-providers/kind-crossplane/Taskfile.yaml
```diff
@@ -910,6 +2373,11 @@ tasks:
       - task: credentials:refresh
       - kubectl -n crossplane-system rollout restart deploy -l pkg.crossplane.io/revision 2>/dev/null || true
       - kubectl -n crossplane-system rollout status deploy -l pkg.crossplane.io/revision --timeout=120s 2>/dev/null || true
+      # Phase 0: Tear down spoke clusters FIRST, while the hub controllers
+      # (KRO/ACK/Crossplane) are still alive to reconcile their deletion. Doing this
+      # after hub:destroy-addons would remove those controllers and orphan the spokes.
+      - cmd: task spokes:disable-all
+        ignore_error: true
```
punkwalker: This is is a good Idea, but we manage all spoke addons from hub, so we should find a way for cleanup from the hub side instead of disabling addons for spoke by committing to git.

cluster-providers/kind-crossplane/Taskfile.yaml
```diff
@@ -1184,9 +2652,66 @@ tasks:
...
-      - kubectl -n crossplane-system wait --for=condition=Ready --timeout=1800s platformclusters.platform.gitops.io/{{.HUB_CLUSTER_NAME}}
+      - kubectl -n crossplane-system wait --for=condition=Ready --timeout=3500s platformclusters.platform.gitops.io/{{.HUB_CLUSTER_NAME}}
       - task: destroy-kind
 
+  spokes:seed-provider-identity:
+    desc: >-
+      Seed the bootstrap Crossplane provider IAM roles + EKS Pod Identity
+      associations for spoke clusters (same account). The iam/eks providers are
+      createIdentity:false in the registry, so crossplane-base does NOT create
+      their roles; without this seed a spoke's iam/eks providers cannot
```
punkwalker, line 1742: This should happen in the addon creation flow, instead of from task file

docs/platform/cluster-lifecycle.md
punkwalker, line 1: Is this a duplicate file from kiro steering files?

gitops/addons/charts/backstage/templates/install.yaml
```diff
@@ -724,14 +791,35 @@ metadata:
   name: backstage
   namespace: backstage
   annotations:
+{{- if eq (default "domain" .Values.exposure_mode) "cloudfront" }}
+    alb.ingress.kubernetes.io/target-type: ip
+    alb.ingress.kubernetes.io/listen-ports: '[{"HTTP":80}]'
+    alb.ingress.kubernetes.io/healthcheck-path: /backstage/health/v1/readiness
+    alb.ingress.kubernetes.io/transforms.backstage: |
+      [{"type":"url-rewrite","urlRewriteConfig":{"rewrites":[{"regex":"^\\/backstage\\/?(.*)$","replace":"/$1"}]}}]
+{{- else }}
```
punkwalker: Isn't backstage is available at "/backstage" regardless if the domain name is CloudFront or otherwise?

gitops/addons/charts/devlake/templates/setup-workflow.yaml
```diff
@@ -35,6 +35,10 @@ metadata:
 spec:
   serviceAccountName: devlake-mysql-setup-sa
   entrypoint: mysql-setup
+  # NOTE: no ttlStrategy. This Workflow is a sync-wave -1 resource tracked by
```
punkwalker, line 38: Is there a way to configure Argo CD to not behave like this?

gitops/addons/charts/grafana/templates/ingress.yaml
```diff
@@ -4,25 +4,32 @@ metadata:
...
+    alb.ingress.kubernetes.io/healthcheck-path: /grafana/api/health
+    alb.ingress.kubernetes.io/success-codes: "200-302"
+    alb.ingress.kubernetes.io/listen-ports: '[{"HTTP":80}]'
+{{- else }}
+    alb.ingress.kubernetes.io/target-type: ip
+    alb.ingress.kubernetes.io/healthcheck-path: /grafana/api/health
+    alb.ingress.kubernetes.io/success-codes: "200-302"
+    alb.ingress.kubernetes.io/listen-ports: '[{"HTTPS":443}]'
+    alb.ingress.kubernetes.io/ssl-redirect: '443'
+{{- end }}
```
punkwalker: Should this if/else just apply to the listen-ports and ssl-redirect annotations?

gitops/addons/charts/ingress-class-alb/templates/ingressclass.yaml
```diff
@@ -10,6 +10,9 @@ spec:
   group:
     name: platform
   scheme: internet-facing
+{{- if .Values.loadBalancerName }}
+  loadBalancerName: {{ .Values.loadBalancerName | quote }}
+{{- end }}
```
punkwalker, line 15: Why do we need loadBalancer name?

gitops/addons/charts/keycloak/templates/keycloak-config.yaml
punkwalker: We will simplify this as per @shapirov103 proposal for splitting this job to app specific KRO resources

gitops/addons/configs/backstage/values.yaml
```diff
@@ -1,7 +1,7 @@
 # Backstage configuration
 # Override these in overlays/environments/<env>/backstage/values.yaml
 
-backstage_image: "ghcr.io/backstage/backstage:latest"
+backstage_image: "public.ecr.aws/seb-demo/backstage:cnoe"
```
punkwalker, line 4: Can we use CNOE image here?

config.yaml
punkwalker: Got it. Can you check the AI-on-EKS inference charts, they have a mechanism for downloading the models to S3. 
Ref: https://github.com/awslabs/ai-on-eks-charts/tree/main/charts/inference-charts

config.yaml
```diff
@@ -49,3 +49,10 @@ identityCenter:
 # EKS ArgoCD Capability
 argocdCapability:
   name: "argocd"
+
+# Console admin role name (used by Backstage templates and added as cluster admin if specified)
+adminRoleName: "<console-role-name>"       # e.g. WSParticipantRole
+
```
punkwalker, line 71: Can you update config schema.json? so that we have some validation locally?

applications/rust/deployment/templates/kro/application.yaml
allamand: It's a place holder replaced by ses if you prefer I can change it

cluster-providers/kind-crossplane/manifests/argocd/create-capability.yaml
allamand: Oh cool yes

cluster-providers/kind-crossplane/Taskfile.cloudfront.yaml
allamand: Why should they need a 2 vpc for this ? Does it work ? Ingress from pods in hub vpc register on separate cloudfront vpc ?

cluster-providers/kind-crossplane/Taskfile.yaml
allamand: I did this in the first place but had many issues with inside jobs. We can still try to improve this later if needed

cluster-providers/kind-crossplane/Taskfile.yaml
allamand: Yes that can be moved in a separate taskfile for the pre-requisite fulfillment, then new section in the config file with required informations

cluster-providers/kind-crossplane/Taskfile.yaml
allamand: This is chicken and eggs. Thoses roles needs to be created for automation to work. But thoses are pre-requisites also, can be moved in a pre-requisite task

cluster-providers/kind-crossplane/Taskfile.yaml
allamand: Yes this one is specific for workshop can move it elsewhere

cluster-providers/kind-crossplane/Taskfile.yaml
allamand: The disable-all is part of the delete script

cluster-providers/kind-crossplane/Taskfile.yaml
allamand: I had very bad results with hooks in the past open to discussion

cluster-providers/kind-crossplane/Taskfile.yaml
allamand: Ok this push for a dedicated workshop taskfile as a pattern

cluster-providers/kind-crossplane/Taskfile.yaml
allamand: This one is tricky as to avoid bad production errors like cluster deletion, we intonionally configured it do not delete AWS ressources when we delete objects for safety. So when we really want to delete something this is an explicit path that requires more actions

cluster-providers/kind-crossplane/Taskfile.yaml
allamand: Need to double check but I guess it is same chicken and eggs problem we need thoses role for addons to work as expected, so can't be done in the addons

docs/platform/cluster-lifecycle.md
allamand: Yes this is more docs for users, maybe redundant, not needed ?

allamand: Steering are more docs for agents

gitops/addons/charts/backstage/templates/install.yaml
allamand: Good point I think yes

gitops/addons/charts/grafana/templates/ingress.yaml
allamand: Yes it can be more narrowed

gitops/addons/charts/ingress-class-alb/templates/ingressclass.yaml
allamand: For adopting existing LB

gitops/addons/charts/keycloak/templates/keycloak-config.yaml
allamand: Can you tell more about this ?

gitops/addons/configs/backstage/values.yaml
allamand: This is cnoe image + minor patches 

For now the patches are still needed and it is built from the backstage/Dockerfile

config.yaml
allamand: Yes and we should include also the pre-requisite conf for cloudfront then also

gitops/addons/charts/devlake/templates/setup-workflow.yaml
allamand: i don't know..

allamand:
@elamaran11 @punkwalker @shapirov103 p’ease look at this new version. All is almost working, we have a dedicated workshop path, it uses different git overlay so we can just use GitHub as mainstream and other git to customize. Happy to chat about it

allamand:
Updated all the recommendations, and tested it for workshop in workshop studio. all is working. Workshop is a separate entry point. Gitlab is managed apart. All good from my side

gitops/addons/charts/keycloak/templates/ingress.yaml
```diff
@@ -1,3 +1,7 @@
+{{- /* insecure: HTTP-only ingress, no cert (consumer terminates TLS upstream).
+       Explicit .Values.global.insecure wins; falls back to the legacy
+       exposure_mode == "cloudfront" toggle during the transition. */ -}}
+{{- $insecure := or .Values.global.insecure (eq (default "domain" .Values.global.exposure_mode) "cloudfront") -}}
```
shapirov103: This still talks about cloudfront - should be removed. Probably a sweep over the PR to make sure references to cloudfront at localized to the workshop folder

shapirov103 (changes requested):
Very cool, getting there!

I stared with comment: https://github.com/aws-samples/appmod-blueprints/pull/709/changes#r3571491629

Here is review based on arch constraints that we agreed upon. Please check the section Leaks to remove for constraint #1


## Architecture review: CloudFront exposure must not leak into the solution

Reviewed against three constraints: (1) no CloudFront-specific logic in the solution — all workshop customization localized to `workshop/`; (2) workshop provisioning is a private case of solution instantiation; (3) the solution supports two modes — an externally-provisioned domain and an insecure mode (for services like Keycloak).

| Constraint | Status |
|---|---|
| #3 — two modes (external domain + insecure) | ✅ Satisfied |
| #2 — workshop = private case of solution instantiation | ✅ Satisfied in structure |
| #1 — no CloudFront leak; workshop specifics localized to `workshop/` | ⚠️ Partially violated |

### ✅ What's right
- The core abstraction is generic and correct: `config.schema.json` defines `domain` (consumer-provisioned — "the platform never creates it"), `insecure`, and `certificateArn`. Ingress templates switch on `.Values.global.insecure` (`keycloak/templates/ingress.yaml:4`, `backstage/templates/install.yaml:790`, `argo-workflows/templates/ingress.yaml:3`). A consumer setting `insecure: true` gets HTTP mode without ever naming CloudFront. **Constraint #3 met.**
- `workshop/` is the CloudFront consumer: `workshop/Taskfile.yaml` provisions the distribution and writes `private/cloudfront-domain` / `private/gitlab-cloudfront-domain`. The generic provider documents spoke/overlay/teardown as consumer concerns. **Constraint #2 met structurally.**

### ⚠️ Leaks to remove for constraint #1
1. **`gitops/addons/registry/core.yaml:211-216`** — `ingress-class-alb` `scheme: internal` and `loadBalancerName` are gated on `exposure_mode == "cloudfront"`. CloudFront-topology logic (VPC origin ⇒ internal ALB) keyed to the consumer's edge name in the core registry. Make it a generic knob (drive from `insecure`, or an explicit `albScheme`).
2. **Legacy `exposure_mode == "cloudfront"` shim in ingress templates** (keycloak, backstage, argo-workflows — comments say "legacy … during transition") OR'd with `insecure`. Collapse to `insecure` only so the string `"cloudfront"` disappears from solution charts.
3. **`gitops/overlays/environments/control-plane/kargo/values.yaml:1`** — self-described "cloudfront variant (HTTP:80 behind the CloudFront origin)" living in the **base** solution overlays. Belongs in the workshop `fleet-config` overlay.
4. **`cluster-providers/kind-crossplane/Taskfile.yaml`** embeds workshop/IDE specifics that already exist in `workshop/Taskfile.yaml`:
   - `EXPOSURE_MODE` derives the literal `"cloudfront"` from `insecure`.
   - reads `private/cloudfront-domain` / `private/gitlab-cloudfront-domain` (lines 794, 951, 1102, 1425).
   - `peeks-ide` CFN stack + `SharedRole` + Playwright inside `idc:configure` (1434-1461); `IDE_PASSWORD` (1009); "git_token seeded by CDK" (1086).
   - generic `install:` calls `install:phase1-playwright-prefetch` (193-204) — Playwright is only needed for the workshop's IDC SAML automation.
   
   `idc:configure` and the cloudfront-domain reads are **duplicated** in `workshop/Taskfile.yaml` (their correct home), so the copies here are the leak to delete.

### Doc drift (not a code leak)
`.kiro/steering/troubleshooting.md` references `cluster-providers/common/Taskfile.cloudfront.yaml`, `install:phase1-cloudfront`, and `create-alb` — none exist on the branch. The generic install does **not** invoke CloudFront provisioning; please update or remove those sections.

### Bottom line
The two-mode design is sound and domain-mode installs work (the CloudFront references are gated fallbacks). To fully satisfy "no CloudFront logic in the solution," please (a) collapse `exposure_mode=="cloudfront"` to `insecure`, (b) make ALB `scheme` generic in `core.yaml`, (c) move the Kargo cloudfront overlay + the `cloudfront-domain`/`peeks-ide`/`IDE_PASSWORD`/Playwright logic into `workshop/`, and (d) fix the stale steering references.

shapirov103 (approved):
LGTM  - thank you for the great work!

@punkwalker I see maybe one or two residual comments that may not have been addressed or are deferred. Please check it out.

punkwalker (approved):

PR #713, opened 2026-06-02
fix: increase EKS cluster creation timeout to 60m

allamand:
## Problem

In fresh accounts (e.g. Workshop Studio sandboxes), EKS Auto Mode clusters can take longer than the Terraform AWS provider default 30m create timeout to become `ACTIVE`. When the waiter times out, `cluster/deploy.sh` retries `terraform apply`, but the clusters are still `CREATING` and the retry fails with:

```
ResourceInUseException: Cluster already exists with name: peeks-hub
```

This causes intermittent `deployment_failed` on the clusters stack.

## Fix

Set an explicit `timeouts` block on the EKS module so the create waiter allows up to 60m:

```hcl
timeouts = {
  create = "60m"
  update = "60m"
  delete = "30m"
}
```

Note: the correct input variable for terraform-aws-modules/eks/aws **v21** is `timeouts` (not `cluster_timeouts`). Validated with `terraform validate`.

## Why only the timeout

We also prototyped a defensive retry path (wait for CREATING→ACTIVE, then `terraform state rm` + `import` to reconcile a half-written state). Testing showed that importing only `aws_eks_cluster` in isolation breaks the EKS module (`Error: Attempt to index null value` on dependent resources like the OIDC provider). The module is not designed for partial imports, so that approach is unreliable and was dropped. Increasing the timeout prevents the half-written state from happening in the first place, which is the robust fix.

## Testing

- `terraform validate` passes.
- Verified the failure mode and root cause end-to-end on Workshop Studio test events.

### DIFF
--- a/platform/infra/terraform/cluster/deploy.sh
+++ b/platform/infra/terraform/cluster/deploy.sh
@@ -119,6 +119,32 @@ main() {
   
   # Apply Terraform configuration
   log "Applying clusters stack..."
+
+  # Helper: wait for any CREATING clusters to reach a terminal state before retrying.
+  # Without this, a retry after a Terraform timeout hits ResourceInUseException because
+  # the clusters are still being created by AWS in the background.
+  wait_for_clusters() {
+    log "Waiting for in-progress EKS clusters to reach a terminal state before retrying..."
+    local cluster_names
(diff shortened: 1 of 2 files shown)


csantanapr (approved):
LGTM

hmuthusamy (approved):
LGTM

PR #529, opened 2026-03-13
Feature/agent platform

allamand:
*Issue #, if available:*

*Description of changes:*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #721, opened 2026-06-05
feat: enable selfHeal on ApplicationSets for automatic drift correction

allamand:
Enables `selfHeal: true` on the `addons` and `clusters` ApplicationSets.

## Why

With `selfHeal: false` (current default), drifted resources accumulate silently and critical resources like NodePools can remain undeployed after sync failures. This caused a cascading issue where:

1. `platform-manifests-bootstrap` sync failed (webhook down)
2. NodePools never applied → no nodes → all pods Pending
3. Manual intervention required every time

## What changes

- `gitops/fleet/bootstrap/addons.yaml`: `selfHeal: false` → `true`
- `gitops/fleet/bootstrap/clusters.yaml`: `selfHeal: false` → `true`

## Risk

Kyverno CRDs may show reconciliation loops due to cosmetic diffs. If this becomes an issue, add `ServerSideDiff` + `RespectIgnoreDifferences` annotations to the kyverno app specifically.

### DIFF
--- a/gitops/fleet/bootstrap/addons.yaml
+++ b/gitops/fleet/bootstrap/addons.yaml
@@ -42,7 +42,7 @@ spec:
         name: '{{.name}}'
       syncPolicy:
         automated:
-          selfHeal: false
+          selfHeal: true
           allowEmpty: true
           prune: false
         retry:
--- a/gitops/fleet/bootstrap/clusters.yaml
+++ b/gitops/fleet/bootstrap/clusters.yaml
@@ -54,7 +54,7 @@ spec:
         name: '{{.name}}'
       syncPolicy:
         automated:
-          selfHeal: false
+          selfHeal: true
           allowEmpty: true
           prune: true
         retry:


shapirov103 (approved):
LGTM

hmuthusamy (approved):
LGTM

allamand:
Closing this as largely superseded by #529.

The intended change here has effectively already landed:

- **Addons ApplicationSet:** `selfHeal: true` is now set in `gitops/bootstrap/addons.yaml` on `main` (the file moved from `gitops/fleet/bootstrap/addons.yaml` during the #529 restructure), with a comment documenting the same rationale as this PR — auto-converge to Git, safe alongside `prune: false`.

- **Cluster ApplicationSets:** the old `gitops/fleet/bootstrap/clusters.yaml` was split into `gitops/bootstrap/clusters-crossplane.yaml` and `gitops/bootstrap/clusters-kro.yaml`. Both currently keep `selfHeal: false` **intentionally** — the inline comments describe this as a deliberate guardrail so ArgoCD never auto-reverts/prunes live cluster provisioning resources.

Because the files this PR targets no longer exist, it shows as a modify/delete conflict. Rather than mechanically re-applying `selfHeal: true` to the new cluster files — which would override the guardrail decision made in #529 on the highest-risk ApplicationSets — I'm closing this.

If we do want `selfHeal` enabled on the cluster-provisioning ApplicationSets, let's do that as a separate, deliberate PR against `clusters-crossplane.yaml` / `clusters-kro.yaml` so it gets reviewed on its own merits.

PR #753, opened 2026-06-23
feat(kro): add protectFromDeletion parameter to EksCluster and Vpc RGDs

allamand:
## Problem

SpringClean deletes EKS clusters and NAT Gateways managed by kro because manually-applied `auto-delete: no` tags are overwritten by kro reconciliation.

## Solution

Adds a new optional schema field `protectFromDeletion` (default: `"true"`) to both the EksCluster and Vpc RGDs. This propagates an `auto-delete` tag via CEL ternary:

```yaml
auto-delete: ${schema.spec.protectFromDeletion == "true" ? "no" : "yes"}
```

**Default behavior**: all clusters and VPCs are protected (opt-out model).

To allow cleanup, set `protectFromDeletion: "false"` in the EksCluster/Vpc instance.

## Changes

- `rg-eks.yaml`: added schema field + `auto-delete` tag on EKS Cluster
- `rg-vpc.yaml`: added schema field + `auto-delete` tag on VPC and NAT Gateway

## Testing

Existing instances don't need changes — the new field defaults to `true`, so next ArgoCD sync will tag all clusters/VPCs with `auto-delete: no` automatically.

### DIFF
--- a/gitops/addons/charts/kro/resource-groups/manifests/eks/rg-eks.yaml
+++ b/gitops/addons/charts/kro/resource-groups/manifests/eks/rg-eks.yaml
@@ -29,6 +29,7 @@ spec:
       git_username: string | default="user1"
       aws_partition: string | default="aws"
       aws_dns_suffix: string | default="amazonaws.com"
+      protectFromDeletion: string | default="true"
       network:
         vpcID: string
         subnets:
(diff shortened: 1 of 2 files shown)


allamand:
Trying to see if a dynamic annotation merge feature could be done in kro

PR #770, opened 2026-07-06
fix(kind-crossplane destroy): run hub:destroy-addons in-process; broaden orphan sweep

elamaran11:
## Problem

Three gaps in `kind-crossplane` `task destroy` leave orphaned AWS resources that block a clean rebuild:

### 1. `hub:destroy-addons` never runs (root cause)
It was invoked via `cmd: task hub:destroy-addons`, which spawns a nested `task` binary through gosh and re-resolves `CONFIG_FILE` relative to the provider dir (`cluster-providers/kind-crossplane/config.local.yaml`) — a path that does not exist:

```
Error: open .../cluster-providers/kind-crossplane/config.local.yaml: no such file or directory
task: [kind-crossplane:destroy] task hub:destroy-addons ... exit status 1
```

Because the step is `ignore_error: true`, the failure is silent and **hub addons are never torn down**, leaving ALBs, AMP scrapers, and addon-created security groups that block VPC deletion. The `argocd:delete-capability` call three lines below already documents this exact pitfall and uses `task:` — this applies the same fix to line 914.

### 2. IAM safety-net misses most roles/policies
Phase 6 deleted only 7 hardcoded hub roles + 3 hub policies, missing the agentcore/bifrost/AMP/Grafana/Keycloak/langfuse/Otel hub roles **and all spoke roles + policies** (~20 roles + 11 policies orphaned in testing).

### 3. Target groups + addon secrets never cleaned
LBC target groups outlive their deleted load balancers and stay pinned to the VPC (blocking VPC delete), and `hub/langfuse-otel` / `hub/langfuse-otel-keys` secrets survived the fixed `config/keycloak/secrets` list.

## Fix

1. `cmd: task hub:destroy-addons` → `task: hub:destroy-addons` (runs in-process, inherits `CONFIG_FILE`).
2. Sweep IAM roles/policies by `<cluster>-` name prefix (hub + spokes from config), including inline policies and non-default policy versions.
3. Add per-VPC target-group deletion; sweep Secrets Manager by `<cluster>/` prefix.

## Testing

Reproduced all three on a live hub+2-spoke teardown, applied the fixes, then ran a full `task destroy CLEANUP_KIND=true` + `task install` rebuild end-to-end — hub and both spokes recreated ACTIVE with no orphaned VPCs, IAM, target groups, or secrets left behind.

### DIFF
--- a/cluster-providers/kind-crossplane/Taskfile.yaml
+++ b/cluster-providers/kind-crossplane/Taskfile.yaml
@@ -910,8 +910,13 @@ tasks:
       - task: credentials:refresh
       - kubectl -n crossplane-system rollout restart deploy -l pkg.crossplane.io/revision 2>/dev/null || true
       - kubectl -n crossplane-system rollout status deploy -l pkg.crossplane.io/revision --timeout=120s 2>/dev/null || true
-      # Phase 1: Clean hub addons (continue on failure so remaining phases still run)
-      - cmd: task hub:destroy-addons
+      # Phase 1: Clean hub addons (continue on failure so remaining phases still run).
+      # Use `task:` (not `cmd: task ...`) so the call runs in-process and inherits
+      # CONFIG_FILE. Spawning a nested `task` binary via gosh re-resolves CONFIG_FILE
+      # relative to this provider dir (cluster-providers/kind-crossplane/config.local.yaml),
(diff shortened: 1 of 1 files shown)


hmuthusamy (approved):
LGTM

PR #776, opened 2026-07-09
feat: Enable Backstage on hub cluster

hmuthusamy:
*Issue #, if available:*

*Description of changes:*
Enable backstage

### DIFF
--- a/gitops/overlays/environments/control-plane/enabled-addons.yaml
+++ b/gitops/overlays/environments/control-plane/enabled-addons.yaml
@@ -41,7 +41,7 @@ enabledAddons:
   kro_manifests_hub: false
   multi_acct: false
   platform_manifests: false
-  backstage: false
+  backstage: true
   kubevela: false
   devlake: false
   gitlab: false


gitops/overlays/environments/control-plane/enabled-addons.yaml
```diff
@@ -41,7 +41,7 @@ enabledAddons:
   kro_manifests_hub: false
   multi_acct: false
   platform_manifests: false
-  backstage: false
+  backstage: true
   kubevela: false
```
elamaran11, line 45: Why is this false

elamaran11 (commented):
Once comment

gitops/overlays/environments/control-plane/enabled-addons.yaml
hmuthusamy: it was disabled, may be from beginning we need to set to true to enable

elamaran11: Lets do it

elamaran11 (approved):

PR #780, opened 2026-07-10
Standardize per-cluster Crossplane env-config (+ XPodIdentity consumer)

shapirov103:
## Summary

Every cluster (hub + all spokes, including future ones) now exposes a cluster-scoped Crossplane `EnvironmentConfig` named **`env-config`** carrying its ambient metadata — `clusterName`, `region`, `vpcId`, `privateSubnetIds`, `publicSubnetIds`, `clusterSecurityGroupId` — as a single source of truth for downstream Compositions. Renames the old `vpc-config` producer to `env-config`.

## How it works

Each cluster's provisioner writes the VPC facts as **annotations on that cluster's ArgoCD cluster-secret**, and a **plain `EnvironmentConfig` manifest** (addon chart) renders `env-config` from them. Fully declarative GitOps — **no Job, no runtime AWS calls, no Pod Identity, no terraform** — and self-healing via ArgoCD.

- **Spokes** — the `platform-cluster` Composition (which reconciles the spoke `PlatformCluster`s on the hub) publishes each subnet id + cluster SG to XR status and writes the annotations, reusing the `Object` that already emitted `aws_vpc_id`.
- **Hub** — created by the Composition on the ephemeral kind bootstrap cluster, so the Composition can't reach the hub's real cluster-secret. The kind provider's `hub:seed-secret` task (the hub's actual provisioner) stamps the same annotations, discovered via AWS CLI during bootstrap.
- **Enablement** — gated on the existing `enable_crossplane` label (env-config is Crossplane-intrinsic; spoke enablement stays consumer-owned), so it lands wherever Crossplane runs with zero consumer wiring.

## Also included

- **`XPodIdentity` abstraction** (XRD + Composition) — the first consumer: callers pass only a `serviceAccount` + `namespace`, and the Composition resolves `clusterName`/`region` from `env-config`, demonstrating the pattern.

## Removed

- Old hub-only `vpc-config` producer (`init-env-config.yaml` + its `external-secret.yaml`).
- The discredited self-discovery Job + its RBAC + `envconfig` Pod Identity (dead code).

## Validation

Live on `peeks-hub`, `spoke-dev`, `spoke-prod` (us-west-2): `env-config` present on all three with the complete field set. Spokes converged automatically after sync; the pre-existing hub was migrated once for `publicSubnetIds` (fresh installs get it from the Taskfile). `helm template` + `yq` clean.

## Reference

Design, rationale, and rejected alternatives: **`docs/env-config-design.md`**.


hmuthusamy (approved):
LGTM. Need to be tested from fresh install though

PR #784, opened 2026-07-15
fix(gitops): remove CloudFront-specific leaks from the generic solution

allamand:
## Summary

Addresses [shapirov103's architecture review](https://github.com/aws-samples/appmod-blueprints/pull/709#pullrequestreview-4685709216) on PR #709 (`CHANGES_REQUESTED`): removes CloudFront-specific logic leaked outside `workshop/` into the generic solution.

Reviewed against the three agreed constraints: (1) no CloudFront-specific logic in the solution — all workshop customization localized to `workshop/`; (2) workshop provisioning is a private case of solution instantiation; (3) the solution supports two modes — an externally-provisioned domain and an insecure mode.

## Changes

- **`core.yaml`**: `ingress-class-alb` scheme/loadBalancerName now driven by the `insecure` annotation (already written to the cluster secret) instead of `exposure_mode==cloudfront`.
- **Wired `insecure` end-to-end** into 5 registries (`platform`, `ml`, `security`, `observability`, `gitops`) — it was never actually connected before this fix, only `exposure_mode` was.
- **Collapsed the `exposure_mode==cloudfront` legacy shim** in 8 templates (keycloak, backstage, argo-workflows x2, grafana, jupyterhub, kubeflow, gitlab) down to `insecure` only.
- **Moved the Kargo ALB ingress override** out of the generic solution (was an orphaned, never-loaded file) into `workshop/overlay/`, seeded into the workshop's fleet-config repo by `workshop/Taskfile.yaml` — consistent with how any consumer customizes their own environment via their own fleet-config repo.
- **`cluster-providers/kind-crossplane`**: removed `idc:configure` (duplicated by `workshop/Taskfile.yaml`'s more robust version, never actually called in this provider's install pipeline); removed `exposure_mode`.
- **`cluster-providers/kind-kro-ack`**: generalized `hub:wait-for-cloudfront-domain` into `hub:wait-for-domain` (real functionality, not a duplicate — just overly CloudFront-specific naming); removed `exposure_mode`.
- **Renamed** `private/cloudfront-domain` → `private/async-domain` and `private/gitlab-cloudfront-domain` → `private/git-domain` (`GITLAB_CF_DOMAIN` → `GIT_DOMAIN`) across both providers and the workshop, since neither concept is CloudFront-specific to the generic solution.
- **Generalized the `insecure` flag onto hub-side KRO RGD ingresses** (`ray-service`, `cicd-pipeline`, `appmod-service`) — they were hardcoded to HTTP:80/`internet-facing` regardless of platform exposure mode. Wired end-to-end for `ray-service` (cpu/gpu/trainium) and `cicd-pipeline` via Backstage (System entity → scaffolder templates → KRO skeletons → RGD schema, CEL ternary on the ALB annotations). `appmod-service` schema/RGD updated but has no Backstage scaffolder (instances are static files edited manually) — not wired further.
- Fixed `.kiro/steering/troubleshooting.md` stale references to files/tasks that no longer exist.

## Verification

- `helm lint`/`helm template` on all touched charts, in both `insecure=true` and `insecure=false` modes.
- `yq` validation on the 3 modified KRO RGDs — caught and fixed an unquoted CEL ternary expression that broke YAML parsing (the ternary's `:` was read as a second YAML key-value separator).
- `task --list-all` on all 3 Taskfiles (2 providers + workshop).
- `git grep` clean of `exposure_mode`/`cloudfront` residuals outside legitimate comments (CloudFront cited only as an example of a consumer's edge, never as hardcoded logic).

## Companion change

A matching rename (`private/gitlab-cloudfront-domain` → `private/git-domain`) was applied to `cdk/resources/bootstrap.sh` in the separate CDK infra repo (the actual writer of that file), on branch `feat/pr-409`.


cluster-providers/kind-kro-ack/Taskfile.yaml
```diff
@@ -895,12 +899,15 @@ tasks:
...
-        # is correctly set even when hub:seed runs shortly after the CF domain is written.
-        RUNTIME_DOMAIN=$(cat {{.ROOT_DIR}}/private/cloudfront-domain 2>/dev/null | tr -d '[:space:]')
+        # time, before the private/async-domain artifact exists). This ensures
+        # ingress_domain_name is correctly set even when hub:seed runs shortly after the
+        # domain is written.
+        RUNTIME_DOMAIN=$(cat {{.ROOT_DIR}}/private/async-domain 2>/dev/null | tr -d '[:space:]')
         [ -z "$RUNTIME_DOMAIN" ] && RUNTIME_DOMAIN=$(yq '.domain // ""' {{.CONFIG_FILE}} 2>/dev/null | tr -d '[:space:]')
+        # Last-resort fallback specific to CloudFront consumers: if neither private/async-domain
+        # nor config.domain resolved anything, check for a matching CloudFront distribution.
         [ -z "$RUNTIME_DOMAIN" ] && RUNTIME_DOMAIN=$(aws cloudfront list-distributions --region {{.AWS_REGION}} \
```
shapirov103: I realize this was not modified by this PR but rather was carried over. However, this is clearly ckoud front specific APIs in the solution task file. If needed, should be moved over to the workshop folder as well.

shapirov103 (changes requested):
Looks awesome, please see my remaining comment.

cluster-providers/kind-kro-ack/Taskfile.yaml
allamand: I think that this last fallback is even not needed and can be removed

PR #787, opened 2026-07-23
feat(karpenter): self-managed Karpenter addon alongside EKS Auto Mode

punkwalker:
## Summary

Adds a **self-managed Karpenter** addon that runs **alongside EKS Auto Mode**, for dedicated workload pools (e.g. Kata nested-virt sandbox nodes) that Auto Mode's Bottlerocket nodes cannot host. The two coexist safely: each controller only reconciles NodePools whose `nodeClassRef.group` matches its NodeClass API (Auto Mode = `eks.amazonaws.com`, self-managed = `karpenter.k8s.aws`), so neither manages the other's NodeClaims. Verified live on a spoke cluster.

Gated behind `enable_karpenter`. Provisions **only the controller + its IAM** — the `EC2NodeClass` + `NodePool` objects are supplied by the consuming layer (e.g. an agentic `kata-nodepool` addon), not here.

## What's included

Single wrapper chart `gitops/addons/charts/karpenter` + a `karpenter` entry in `registry/core.yaml`:

- **Upstream karpenter `1.14.0` as a Helm dependency** (values nested under `karpenter:`). 1.14.0 is the first release exposing `cpuOptions.nestedVirtualization`, which Kata micro-VMs require.
- **Controller scheduling** — pinned to Auto Mode's built-in `system` nodepool via required `nodeAffinity` `karpenter.sh/nodepool In [system]`. The upstream default (`DoesNotExist`) matches nothing on Auto Mode (every node carries the label) → the controller would stay `Pending`. A subchart table can't be nulled via Helm coalesce, so we replace the `nodeSelectorTerms` instead.
- **Controller pod-identity** — IAM Role (trusted by `pods.eks.amazonaws.com`) + the official Karpenter v1.14 controller policy + a `PodIdentityAssociation`. Uses **EKS Pod Identity, not IRSA**, matching this platform.
- **Node IAM** — EC2-trusted Role + worker/CNI/ECR/SSM managed policies + an `EC2_LINUX` `AccessEntry` so Karpenter-launched nodes join.
- **`node.manageAddons` (default `false`)** — optionally adopt-or-create the `vpc-cni` + `kube-proxy` EKS addons a self-managed node needs (Auto Mode's networking covers only its own nodes). Default off so it never dual-owns addons the `platform-cluster` Composition already manages on managed-node-group clusters; enable only on a pure Auto Mode cluster.
- All Crossplane MRs use `deletionPolicy: Orphan` + `SkipDryRunOnMissingResource` — disabling the addon never deletes a role a live node is still assuming.

## CRD coexistence

The upstream subchart ships CRDs in `crds/` (Helm install-if-absent) and the addons app uses `ServerSideApply`, so Auto Mode's **shared** `karpenter.sh` `nodepools`/`nodeclaims` CRDs are left intact; only the non-shared `karpenter.k8s.aws/ec2nodeclasses` CRD is added.

## Testing

Validated on a live spoke (pure Auto Mode) during development of the coexistence approach:
- self-managed Karpenter provisioned a dedicated tainted node while Auto Mode was undisturbed (5 nodes stayed Ready, its NodePools uneventful);
- NodeClaims partitioned cleanly by `nodeClassRef.group` — zero cross-controller ownership;
- workloads with a required kata label + toleration scheduled onto the Karpenter node only; normal pods stayed on Auto Mode (Karpenter ignored them, its pool being tainted).

Chart renders clean; `node.manageAddons` gate verified (0 Addon MRs off, 2 on); embedded controller policy JSON validates.

### DIFF
--- a/gitops/addons/charts/karpenter/Chart.yaml
+++ b/gitops/addons/charts/karpenter/Chart.yaml
@@ -1,0 +1,18 @@
+apiVersion: v2
+name: karpenter
+description: >-
+  Self-managed Karpenter (alongside EKS Auto Mode) as a single addon: wraps the
+  upstream karpenter controller chart as a dependency (values nested under the
+  `karpenter:` key) and adds the Crossplane-managed IAM it needs — controller
+  pod-identity (Role + official v1.14 controller policy + PodIdentityAssociation)
+  and the node IAM role (+ managed policies + EC2_LINUX access entry) — plus the
+  optional (manageAddons-gated) vpc-cni/kube-proxy EKS addons a self-managed node
+  requires when the cluster has none. One enable_karpenter flip brings up the
+  controller + all its IAM together.
+type: application
+version: 0.1.0
+dependencies:
+  - name: karpenter
+    # First upstream release with cpuOptions.nestedVirtualization (Kata).
+    version: '1.14.0'
+    repository: oci://public.ecr.aws/karpenter
--- a/gitops/addons/charts/karpenter/templates/controller-identity.yaml
+++ b/gitops/addons/charts/karpenter/templates/controller-identity.yaml
@@ -1,0 +1,297 @@
+{{- if .Values.controller.enabled }}
+{{- /*
+Karpenter CONTROLLER pod-identity: IAM Role (pods.eks.amazonaws.com trust) +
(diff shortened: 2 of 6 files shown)


elamaran11 (approved):
LGTM

PR #788, opened 2026-07-24
fix(platform): bump fluent-bit 3.4.9 + fix keycloak/argo-workflows in…

hmuthusamy:
…gress insecure flag

1. Bump aws-for-fluent-bit from 3.4.8 to 3.4.9 to resolve high-severity CVEs (ALAS2023-2026-1986, ALAS2023-2026-1942, CVE-2026-11850).

2. Fix keycloak and argo-workflows ingress templates: the insecure flag was evaluated as a bare Go template conditional, but the ApplicationSet passes the string "false" which is truthy. Changed to explicit eq comparison so insecure=false correctly renders HTTPS listen-ports with the host field set, placing the ingress rules on the 443 listener alongside Langfuse and AgentGateway.

*Issue #, if available:*

*Description of changes:*


elamaran11 (approved):
LGTM

PR #790, opened 2026-07-24
fix(karpenter): set forProvider.region on EKS PodIdentityAssociation …

punkwalker:
…+ AccessEntry

Every eks.aws.upbound.io managed resource requires spec.forProvider.region; the PodIdentityAssociation and AccessEntry templates omitted it, so both failed admission at sync:

  spec.forProvider.region: Required value

.Values.aws.region is already injected (the controller policy uses it throughout) — wire it into both forProvider blocks. The iam.aws.upbound.io MRs (Role/Policy/RolePolicyAttachment) are global and unaffected.

*Issue #, if available:*

*Description of changes:*


hmuthusamy (approved):
LGTM

PR #797, opened 2026-07-29
fix(karpenter): region + explicit enableNetworkPolicy on EKS Addon MRs

punkwalker:
## Problem

The `manageAddons`-gated vpc-cni / kube-proxy `Addon` MRs are missing `spec.forProvider.region`. The `eks.aws.upbound.io` Addon CRD lists `region` in `forProvider.required`:

```
$ kubectl get crd addons.eks.aws.upbound.io -o jsonpath='{...forProvider.required}'
["region"]
```

So setting `node.manageAddons=true` fails admission with `spec.forProvider.region: Required value` and **neither addon is ever created** — the gate is unusable. Same class as the region fixes already landed for PodIdentityAssociation (#790) and AccessEntry (#791).

## Changes

**1. `region: {{ .Values.aws.region }}` on both Addon MRs.**

**2. Explicit `enableNetworkPolicy` in the vpc-cni `configurationValues`.**

These MRs adopt in place with `resolveConflictsOnUpdate: OVERWRITE`, which takes field ownership. Adopting a cluster that already had NetworkPolicy enforcement on would otherwise **silently turn it off**, rendering every `NetworkPolicy` / `ClusterNetworkPolicy` inert — a self-managed sandbox node's egress containment depends on it. Defaults to `"true"` (fail-secure); overridable via `node.addons.vpcCni.enableNetworkPolicy`.

**3. `configurationValues` built with `dict | toJson`** instead of hand-written JSON, so quoting/escaping is always well-formed.

The value is read with `hasKey`+`get` rather than `| default`, because Helm's `default` treats bool `false` as empty and would flip an explicit `false` back to `true`. An overlay may now write either `false` or `"false"`.

It serializes as the **string** `"true"`/`"false"`, matching the format EKS accepts in practice (verified against a live cluster running `{"enableNetworkPolicy":"true"}` with enforcement working).

## Verified

`node.manageAddons` remains **`false` by default** — with it unset, no Addon MRs render at all, so this is inert for every existing cluster.

With `manageAddons=true`, both MRs render `region: us-west-2`, and `configurationValues` is valid JSON with `enableNetworkPolicy` typed `str` across all four override forms:

| Input | Rendered |
|---|---|
| (default / key absent) | `"true"` |
| `enableNetworkPolicy: "false"` | `"false"` |
| `enableNetworkPolicy: false` (bool) | `"false"` |
| overlay replaces `node:` wholesale | `"true"` (fail-secure) |

`helm lint` clean.

## Note for reviewers

The `region` fix is required for the feature to work at all. The `enableNetworkPolicy` addition is a **deliberate judgement call** — it makes the chart take ownership of a security-relevant field it previously left alone. Happy to split it out if you'd prefer the region fix land on its own.

Unchanged in this PR: the vpc-cni toleration remains `{"operator":"Exists"}`, which tolerates every taint. On a cluster with existing Auto Mode nodes, enabling `manageAddons` would therefore schedule `aws-node` onto them too. Worth scoping with a nodeSelector in a follow-up; not made worse here since the default stays off.

### DIFF
--- a/gitops/addons/charts/karpenter/templates/eks-addons.yaml
+++ b/gitops/addons/charts/karpenter/templates/eks-addons.yaml
@@ -12,7 +12,10 @@ GATED behind node.manageAddons (default false): EKS addons are per-cluster
 singletons, so this must be OFF on any cluster where something else owns them
 (e.g. the platform-cluster Composition on managed-node-group clusters) or the two
 owners fight. Enable ONLY on a pure Auto Mode cluster with no existing owner.
-The vpc-cni toleration lets it schedule onto the tainted kata node.
+
+Neither addon needs a tolerations override to reach the tainted self-managed node:
+both DaemonSets already ship `{operator: Exists}`. See the vpc-cni
+configurationValues below for what actually governs placement.
 */ -}}
 apiVersion: eks.aws.upbound.io/v1beta1
 kind: Addon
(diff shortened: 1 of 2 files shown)


punkwalker:
## Update: dropped the tolerations override (and retracting a caveat)

Pushed `2e02e39f`. Review feedback correctly pointed out that the live spoke-dev addon runs **without** any tolerations in its `configurationValues` — so I checked why it works, and the override was redundant.

**Both DaemonSets already tolerate everything.** `aws-node` and `kube-proxy` ship with:

```
key= operator=Exists effect=
```

i.e. a blanket toleration for every taint. Neither needs help reaching a tainted self-managed node.

**What actually governs placement is nodeAffinity**, not tolerations:

```
eks.amazonaws.com/compute-type NotIn [fargate, hybrid, auto]
```

Auto Mode nodes carry `compute-type=auto` and are excluded; a self-managed node leaves it unset and is included. Confirmed live on spoke-dev — `aws-node` runs on the self-managed kata node and on **none** of the 7 Auto Mode nodes.

### Retracting the caveat in the PR description

The description's closing note said enabling `manageAddons` would schedule `aws-node` onto existing Auto Mode nodes, and suggested scoping it with a nodeSelector. **That concern was wrong** — affinity already excludes those nodes regardless of tolerations. I had reasoned from the toleration alone without checking what gates placement. No nodeSelector follow-up is needed.

### Net effect

`configurationValues` now renders exactly what a working cluster already runs:

```json
{"enableNetworkPolicy":"true"}
```

Re-verified after the change: both MRs carry `region: us-west-2`; `kube-proxy` has no `configurationValues` at all; and the override paths still behave (`"false"` → `"false"`, bool `false` → `"false"`, `node:` replaced wholesale → `"true"` fail-secure). `helm lint` clean.

The substance of the PR is unchanged: `region` is what makes `manageAddons` usable at all, and `enableNetworkPolicy` guards against OVERWRITE silently stripping enforcement.

shapirov103 (approved):
LGTM

PR #798, opened 2026-07-29
feat(fleet-secret): optional external overlay repo annotations (layer 5)

punkwalker:
## Problem

The appset-chart **already** supports an external overlay repo. `bootstrap/addons.yaml` reads three annotations into `overlayRepoURLGit*`, and the appset-chart appends a `$overlay` source plus three per-addon value files that layer *after* the in-repo ones.

But nothing ever emitted those annotations, so the feature was unreachable. Unlike `overlay_repo_basepath` (which has an `or` fallback), `overlayRepoURLGit` has none — it stays empty, and the appset-chart's `if $overlayRepoURLGit` guard skips the per-addon overlay layer entirely.

Concretely, today the generated Applications have only two sources:

```
$ kubectl get application karpenter-spoke-dev -n argocd -o yaml
  [0] ref=defaults  repoURL=appmod-blueprints
  [1]               repoURL=appmod-blueprints  path=gitops/addons/charts/karpenter
        VF: $defaults/gitops/addons/configs/karpenter/values.yaml
        VF: $defaults/gitops/overlays/environments/dev/karpenter/values.yaml
        VF: $defaults/gitops/overlays/clusters/spoke-dev/karpenter/values.yaml
```

`$defaults` resolves to the repo that **owns the addon**. So for an addon whose chart + registry entry live here, layers 3–4 are read from *this* repo — a consumer repo wanting to change one value for one cluster has to open a PR here.

And it fails *silently*: every overlay path carries `ignoreMissingValueFiles: true`, so a `values.yaml` placed in the consumer repo is skipped without error and the setting simply never takes effect. That is a genuinely hard failure mode to debug.

## Change

An optional `overlay: {repoURL, revision, basepath}` block that renders the annotations:

```yaml
overlay:
  repoURL: ""        # default OFF
  revision: "main"
  basepath: ""
```

Rendered in **both** templates — `cluster-secret.yaml` (`mode: direct`, spokes) and `external-secret.yaml` (`mode: externalSecret`, hub). Doing only one would have looked correct while silently no-opping for the other cluster type.

Gated on `{{- if .repoURL }}`, so **behaviour is unchanged for every existing cluster**.

## How to use it

**1. Set `overlay` in the fleet-member values** (`fleet/hub/values.yaml`, or `fleet/members/<cluster>/values.yaml`):

```yaml
overlay:
  repoURL: https://github.com/my-org/my-consumer-repo.git
  revision: main
  basepath: gitops/
```

Per-cluster: clusters without it keep reading overlays from the addon's own repo.

**2. Add the override in the consumer repo**, at the path for that addon + cluster:

```yaml
# gitops/overlays/clusters/hub/karpenter/values.yaml   (in the CONSUMER repo)
node:
  manageAddons: true
```

The file takes the **addon chart's** value schema — same shape as `addons/charts/<addon>/values.yaml`.

**3. Verify it resolved.** The Application should now have three sources and `$overlay` value files:

```sh
kubectl get application <addon>-<cluster> -n argocd -o yaml | grep -A5 valueFiles
```

If the `$overlay` entries are missing, the annotation did not reach the cluster secret:

```sh
kubectl get secret <cluster> -n argocd -o jsonpath='{.metadata.annotations.overlay_repo_url}'
```

## Caveats (documented in the README)

- **`$overlay` is a real ArgoCD source.** An unreachable or credential-requiring URL breaks **every** overlay-aware app on that cluster, not just the one being overridden. Register a repo credential first; wire it late in a bootstrap sequence, after the repo exists.
- **`revision` must contain the files**, or the overlay silently no-ops and the value quietly stays at its in-repo default. Update it when the work merges to the default branch.
- **It also repoints the appset-level `overrides.yaml` reads.** Independently of the per-addon layers, `bootstrap/addons.yaml` reads `$overlay/<basepath>overlays/{environments/<env>,clusters/<cluster>}/overrides.yaml`, which fall back to this repo when the annotation is unset. Those take the **registry-entry** schema (addon-keyed, `valuesObject` / `defaultVersion`), *not* chart values — so chart values placed there do nothing.
- **Only the platform appset consumes these annotations.** A consumer repo running its own appset with its own registry files never sets `overlayRepoURLGit`, so its addons keep resolving overlays through layer 4 as before — no double-read.

## Verified

| Case | Result |
|---|---|
| `direct` (spoke), default | 0 annotations |
| `direct` (spoke), `repoURL` set | 3 annotations |
| `externalSecret` (hub), default | 0 annotations |
| `externalSecret` (hub), `repoURL` set | 3 annotations |
| `revision` omitted | defaults to `main` |
| Rendered output, both modes | parses as valid YAML |

`helm lint` clean.

## Docs

`README.md` gains an **External Overlay Repo ("layer 5")** section: the layering model (why `$defaults` is per-addon), the step-by-step walkthrough above with verification commands, and the caveats. The `values.yaml` reference block is updated too.

### DIFF
--- a/platform-charts/fleet-secret/README.md
+++ b/platform-charts/fleet-secret/README.md
@@ -36,6 +36,11 @@ externalSecret:
   annotations: {}          # Extra annotations on the cluster secret
   labels: {}               # Extra labels on the cluster secret
 
+overlay:                   # External overlay repo ("layer 5") -- OPTIONAL, default off
+  repoURL: ""              # Consumer repo that may override platform addon values
+  revision: "main"         # MUST contain the overlay files
+  basepath: ""             # Root holding configs/ + overlays/ (e.g. "gitops/")
+
 enabledAddons: {}          # Populated from enabled-addons.yaml via valueFiles
 ```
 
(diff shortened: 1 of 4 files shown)


shapirov103 (approved):
LGTM

We had imperative set hub:set-overlay-repo before, I think the spokes were missing, so looks solid

PR #796, opened 2026-07-29
feat: enable Managed KRO + ACK capabilities on the hub

elamaran11:
## Enable Managed KRO + Managed ACK capabilities on the hub

Enables the **Managed KRO** and **Managed ACK** EKS Capabilities on the hub, alongside the existing **Managed ArgoCD** capability.

### Context
The `Capability` MRs for KRO / ACK / ARGOCD already exist in the `platform-cluster` Crossplane Composition (provider-aws-eks v2.6.1), gated by `spec.capabilities.<type>.enabled` via `function-cel-filter`, with their IAM roles. This PR **turns KRO + ACK on** for the hub and documents/validates the toggle.

### Changes
- **`config.yaml`** — hub `capabilities: { kro.enabled: true, ack.enabled: true }` (enabled by default; spokes stay opt-in via the XRD default `false`).
- **`config.schema.json`** — adds the `hub.capabilities` schema (kro/ack `enabled` booleans).
- **`docs/EKS-Capabilities-KRO-ACK-Setup.md`** (new) — how to enable/verify/troubleshoot the KRO+ACK capabilities; mirrors the ArgoCD capability doc. **Generic** — no consumer specifics.

### Scope
This PR is **purely capability enablement**. Consumer-specific details (which non-GA controllers a workload self-manages, the ACK GA services list, Flow D / Lambda MicroVM) live with the **consumer** in `sample-open-agentic-platform` (the Flow D PR), not here.

- **KRO** → serves `ResourceGraphDefinition`s.
- **ACK** → manages **GA** AWS resources (iam/s3/rds/…) as Kubernetes CRs. Non-GA controllers are a consumer's self-managed concern.

### Verification
- `config.schema.json` + `config.yaml` parse; the new `hub.capabilities` block validates (the only schema-validation miss is the pre-existing `<prefix>` template placeholder, unrelated).
- Capabilities remain default-disabled on spokes; hub un-gates the KRO/ACK Capability MRs → `kubectl get capability.eks.aws.upbound.io -A` should reach Ready after sync.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #41, opened 2024-10-08
`repoUrl` missing on app-deploy template

elamaran11:
*Issue #, if available:*

*Description of changes:*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #800, opened 2026-07-30
feat: codify KRO + ACK capability IAM roles for the hub

elamaran11:
## Codify KRO + ACK capability IAM roles for the hub

Follow-up to #796 (which enabled the `capabilities.kro/ack` toggle). Adds the **IAM capability roles** those capabilities need, mirroring the existing `argocd-capability-role.yaml`, so a hub bootstrap creates them **declaratively** instead of by hand.

### Why
The platform-cluster Composition creates the KRO/ACK `Capability` MRs when `spec.capabilities.<type>.enabled` is true, but each Capability references an IAM role by external-name (`<cluster>-KROCapabilityRole` / `<cluster>-ACKCapabilityRole`). Those role manifests didn't exist (only ArgoCD's did) — so the roles would have been drift.

### Changes
- **`claims/kro-capability-role.yaml`** — Crossplane IAM `Role` MR, `capabilities.eks.amazonaws.com` trust only (KRO orchestrates in-cluster resources; the ACK controllers make the AWS calls, so KRO needs no AWS perms).
- **`claims/ack-capability-role.yaml`** — IAM `Role` MR + a **least-privilege inline `RolePolicy`** scoped to Flow D's GA resources: `iam` (roles matching `*-microvm-build`/`*-microvm-exec`) + `s3` (buckets matching `*-microvm-artifacts`). No broad managed policies.
- **`kind-crossplane/Taskfile.yaml`** — apply both roles after the ArgoCD role in `hub:seed`, and delete them in teardown (mirrors the ArgoCD wiring).

### Applied to the live hub
These MRs were applied to the running hub (its bootstrap KIND cluster is gone, so the composition couldn't retro-create them) and reconciled **Ready** — adopting the roles via external-name. KRO + ACK capabilities are now **ACTIVE** on the hub (`kubectl get crd | grep -E 'kro.run|services.k8s.aws'` shows `resourcegraphdefinitions.kro.run` + `iam`/`s3` groups). This PR makes that reproducible for future hub builds.

### Consumer
These capabilities back **Flow D — Lambda MicroVM Agent Sandbox** in `sample-open-agentic-platform` (PR #41).

Generated with Claude Code.

### DIFF
--- a/cluster-providers/kind-crossplane/claims/kro-capability-role.yaml
+++ b/cluster-providers/kind-crossplane/claims/kro-capability-role.yaml
@@ -1,0 +1,33 @@
+# IAM Role for the EKS KRO Capability
+# Mirrors argocd-capability-role.yaml. Created by Crossplane before the KRO
+# capability is created. KRO orchestrates in-cluster Kubernetes resources
+# (ResourceGraphDefinitions); the underlying AWS calls are made by the ACK
+# controllers, so this role needs only the capability trust — no AWS permissions.
+apiVersion: iam.aws.upbound.io/v1beta1
+kind: Role
+metadata:
+  name: kro-capability-role
+  namespace: crossplane-system
+  annotations:
+    crossplane.io/external-name: CLUSTER_NAME-KROCapabilityRole
+spec:
+  forProvider:
+    assumeRolePolicy: |
+      {
+        "Version": "2012-10-17",
+        "Statement": [
+          {
+            "Effect": "Allow",
+            "Principal": {
+              "Service": "capabilities.eks.amazonaws.com"
+            },
+            "Action": [
+              "sts:AssumeRole",
+              "sts:TagSession"
+            ]
+          }
+        ]
+      }
+    tags:
+      managed-by: crossplane
+      purpose: eks-kro-capability
--- a/cluster-providers/kind-crossplane/Taskfile.yaml
+++ b/cluster-providers/kind-crossplane/Taskfile.yaml
@@ -435,6 +435,13 @@ tasks:
           {{.MNG_SETS}}
           | kubectl apply -f -
       - sed "s|CLUSTER_NAME|{{.HUB_CLUSTER_NAME}}|g" claims/argocd-capability-role.yaml | kubectl apply -f -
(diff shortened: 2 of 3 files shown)


cluster-providers/kind-crossplane/claims/kro-capability-role.yaml
```diff
@@ -0,0 +1,33 @@
...
+# capability is created. KRO orchestrates in-cluster Kubernetes resources
+# (ResourceGraphDefinitions); the underlying AWS calls are made by the ACK
+# controllers, so this role needs only the capability trust — no AWS permissions.
+apiVersion: iam.aws.upbound.io/v1beta1
+kind: Role
+metadata:
+  name: kro-capability-role
+  namespace: crossplane-system
+  annotations:
+    crossplane.io/external-name: CLUSTER_NAME-KROCapabilityRole
```
hmuthusamy, line 12: @elamaran11  did we mess the $ here?

elamaran11:
No `$` needed here — `CLUSTER_NAME` is a plain literal placeholder, not a shell/Taskfile variable. It's substituted by `sed` at apply time, exactly like the existing `argocd-capability-role.yaml`:

```
# cluster-providers/kind-crossplane/Taskfile.yaml
sed "s|CLUSTER_NAME|{{.HUB_CLUSTER_NAME}}|g" claims/kro-capability-role.yaml | kubectl apply -f -
sed "s|CLUSTER_NAME|{{.HUB_CLUSTER_NAME}}|g" claims/ack-capability-role.yaml | kubectl apply -f -
```

The `{{.HUB_CLUSTER_NAME}}` on the sed *replacement* side is the Taskfile var; the `CLUSTER_NAME` token in the manifest is intentionally bare so `sed` swaps it (e.g. `hub-KROCapabilityRole`). Adding a `$` would actually break the match. This mirrors `argocd-capability-role.yaml` line 9 (`CLUSTER_NAME-ArgoCDCapabilityRole`) verbatim.

Verified live on the hub: the roles reconciled Ready via external-name and both KRO + ACK capabilities are ACTIVE.

hmuthusamy (approved):
LGTM

PR #801, opened 2026-07-30
fix(kargo): use single-line annotation to avoid YAML parse error

hmuthusamy:
The multi-line pipe literal with escaped backslashes in the ALB transforms annotation was being mangled through the ApplicationSet Go template → Helm values pipeline, causing a YAML parse error in the Kargo chart template rendering. Switch to a single-line quoted string with unescaped regex.

*Issue #, if available:*

*Description of changes:*


elamaran11 (approved):
LGTM

PR #802, opened 2026-07-30
fix(kargo): fix argocd.urls format (map not list) and restore URL rew…

hmuthusamy:
…rite

Root cause: argocd.urls changed from a list to a map in Kargo 1.11.0. The chart renders urls as key=value pairs in the configmap and a list caused Go fmt to emit invalid YAML.

Verified with helm template --version 1.11.0 (199 resources, clean).

*Issue #, if available:*

*Description of changes:*


elamaran11 (approved):
LGTM

PR #810, opened 2026-08-17
Merge origin/main into feature/agent-platform

allamand:
## Summary

Merges `origin/main` (main-line CVE fixes, chart version bumps, template fixes) into `feature/agent-platform`, resolving structural and content conflicts from the divergent registry/addons refactor.

## Structural decisions
- **Terraform removed** (`platform/infra/terraform/*`) — feature branch's GitOps-only path confirmed as primary; main's terraform changes not ported.
- **`.kiro/skills/*` relocated** from `hack/.kiro/skills/` to `.kiro/skills/` (93 files), matching main's directory rename.
- **Fleet bootstrap ported** into the registry/ split architecture (`gitops/addons/registry/*.yaml`, `gitops/bootstrap/*.yaml`) — chart version bumps were already up to date; added the `ExternalSecret` `nullBytePolicy` `ignoreDifferences` pattern (~15 charts) and the new `argocd-health-checks` addon entry.
- **3 addons dropped** (`ingress-nginx`, `crossplane-aws`, `flux`) — confirmed intentional, superseded by the registry refactor.

## Content conflicts resolved (11 files)
- `kro/resource-groups/manifests/eks/rg-eks.yaml` (9 hunks) — kept feature's more advanced ArgoCD-managed-Capability model (IAM Role + Capability + AccessEntry) over main's simpler AccessEntry-only approach; ported main's `readyWhen` ACK-sync gating.
- `kro/resource-groups/manifests/appmod-service.yaml` — kept feature's simple metrics block (main's CEL `evaluationCriteria` extension not portable, per `docs/UPGRADE-FROM-MAIN-2026-05.md`).
- `backstage/templates/install.yaml` — kept feature's additions (dynamic catalog, WORKSHOP_GIT_BRANCH, startupProbe path — documented as intentional in `docs/UPGRADE-FROM-MAIN-2026-05.md`).
- `keycloak/templates/keycloak-config.yaml` — ported main's Secrets Manager push logic (was missing at this call site; de-duplicated against an existing copy).
- `grafana-dashboards/templates/dora-dashboards.yaml` — kept feature's pinned DevLake version + datasource-by-name (main's UID-based datasource doesn't match our provisioned datasource).
- `grafana-dashboards/templates/rust-dashboard.yaml` — hybrid fix: correct container/namespace names from main (`rust-microservice`/`team-rust`) + correct platform label convention from feature (`k8s_container_name`/`k8s_namespace_name`). **Needs live validation** — tracked as TODO in `docs/UPGRADE-FROM-MAIN-2026-05.md`.
- `multi-acct/templates/iam-role-selectors.yaml`, `kro/resource-groups/manifests/eks/rg-eks-vpc.yaml`, `backstage/eks-cluster-template/template.yaml` — kept feature (consistency with rest of file / actual directory structure).
- `scripts/validation/backstage-auth.sh` — kept feature's cookie-jar handling and token-refresh strategy throughout.

## Testing
- `helm template` validated on grafana-dashboards, backstage, keycloak, multi-acct charts (all render clean).
- `helm template` validated on the `kro-clusters` chart with the `argoCdCapabilityRoleArn` fallback change.
- Full validation checklist in `docs/UPGRADE-FROM-MAIN-2026-05.md` Phase 5 still needs a live `task install` run against a real cluster (not done here).

## Blocked features / follow-ups
- rust-dashboard.yaml fix needs live Grafana validation (see TODO).
- `.kiro/steering/project.md` still exists at `hack/.kiro/steering/project.md` on `main` — should be cleaned up there directly (separate PR).


allamand:
@shapirov103 i want to fix conflicts for merging https://github.com/aws-samples/appmod-blueprints/pull/529, this PR should help with that, can you review ?

shapirov103 (commented):
One issue I think should be fixed before merge, plus two trivia and one decision to confirm.

## Blocker: the cicd-pipeline self-service path breaks

This PR repoints both scaffolder skeletons from `project: default` to `project: spoke-workloads`, but the AppProject that defines `spoke-workloads` is never applied, so every app scaffolded from the cicd-pipeline template will emit an ArgoCD Application referencing a nonexistent project. ArgoCD rejects those (`Application referencing project spoke-workloads which does not exist`). It fails silently: the template succeeds, the repo is created, only the Application fails afterwards.

Why it is never applied:

- `gitops/bootstrap/root-appset.yaml:23` syncs `path: '{{ .metadata.annotations.addonsRepoBasepath }}bootstrap'` with `recurse: false` (line 26).
- `config.yaml:13` sets `basepath: "gitops/"`, so that resolves to `gitops/bootstrap/`.
- `appprojects.yaml` lives in `gitops/fleet/bootstrap/`, a sibling directory, outside that path. Nothing else in the tree references `fleet/bootstrap` (only historical mentions in `docs/UPGRADE-APPROACH.md`).

Worth noting explicitly, because it is easy to assume this resolves itself on the merge back to main: on `main` today `gitops/fleet/bootstrap/` **is** the live bootstrap directory (it holds main's `addons.yaml`, `clusters.yaml`, `fleet-secrets.yaml`), and it is applied imperatively by Terraform (`platform/infra/terraform/common/variables.tf:143`, "The path of fleet bootstraps in the repo"). This PR deletes both halves of that mechanism: main's bootstrap appsets under `gitops/fleet/bootstrap/` are dropped in favour of `gitops/bootstrap/`, and all 89 terraform files go away. So `appprojects.yaml` is a leftover from main's old layout that survived only because it had nothing to conflict with, and it stays orphaned after the merge back.

Two ways to close it:

1. Revert the two skeleton manifests to `project: default` and delete `gitops/fleet/bootstrap/appprojects.yaml`. Restores today's working behaviour.
2. Keep the scoping: move `appprojects.yaml` directly into `gitops/bootstrap/` (it must sit at the top level, `recurse: false`), and fix the destination glob noted inline below.

## Trivia

- `gitops/addons/clusters/peeks-hub/addons/backstage/values.yaml` is a 0-byte file at a path no appset reads (the real overlay path is `overlays/clusters/<cluster>/<addon>/values.yaml`). It also bakes the `peeks-hub` cluster name into the generic solution. Suggest deleting.
- `gitops/addons/charts/keycloak/templates/keycloak-config.yaml`: `echo "✓ Keycloak configuration completed successfully!"` now appears twice. Cosmetic merge artifact.

## Decision to confirm, not a defect

The merge back into main deletes 89 terraform files, including `platform/infra/terraform/hub-config.yaml` and the `deploy.sh` / `destroy.sh` path. The PR body says this is intentional, and I agree it follows from the GitOps-only direction. Flagging it only because it makes the merge back a bootstrap-path cutover rather than a file-level merge: any cluster still pointing at the old fleet basepath loses its bootstrap, so it needs to be a conscious sequencing decision.

## Also checked, clean

The `hack/.kiro/` change is 105 renames plus a single real deletion (`hack/.kiro/steering/project.md`), and it converges on the layout `main` already ships, so no content is lost. The stale `hack/.kiro` references in `README.md` and `scripts/sync-apex-skills.sh:53` are pre-existing on the feature branch and untouched here, worth a separate cleanup. Leftover `flux` / `crossplane-aws` mentions for the dropped addons are comments and README only. `readyWhen` gating on `rg-eks.yaml` looks right; only note is that gating on the ExternalSecret `Ready` condition adds a new place spoke provisioning can appear to stall on a slow ESO sync.

platform/backstage/templates/cicd-pipeline/skeleton/manifests/argo-app-dev.yaml
```diff
@@ -17,7 +17,7 @@ metadata:
   finalizers:
   - resources-finalizer.argocd.argoproj.io
 spec:
-  project: default
+  project: spoke-workloads
```
shapirov103, line 20: This changes from `project: default` to `project: spoke-workloads`, but the `spoke-workloads` AppProject is never created on the cluster (see the review summary: `gitops/fleet/bootstrap/appprojects.yaml` sits outside the `gitops/bootstrap/` path that `root-appset.yaml` syncs with `recurse: false`).

Every app scaffolded from this template will produce an ArgoCD Application that ArgoCD refuses with `Application referencing project spoke-workloads which does not exist`. Note every appset in `gitops/bootstrap/` uses `project: default`.

Suggest reverting to `project: default`, or moving the AppProject into `gitops/bootstrap/` so it is actually applied.

platform/backstage/templates/cicd-pipeline/skeleton/manifests/argo-app-prod.yaml
```diff
@@ -17,7 +17,7 @@ metadata:
   finalizers:
   - resources-finalizer.argocd.argoproj.io
 spec:
-  project: default
+  project: spoke-workloads
```
shapirov103, line 20: Same as the dev manifest: `spoke-workloads` is not created anywhere that gets applied, so scaffolded prod Applications will be rejected by ArgoCD. This one is the production path, so the failure is more visible.

gitops/fleet/bootstrap/appprojects.yaml
```diff
@@ -0,0 +1,46 @@
...
+    argocd.argoproj.io/sync-wave: "0"
+spec:
+  description: Workload deployments to spoke clusters with scoped access
+  sourceNamespaces:
+  - argocd
+  sourceRepos:
+  - '*'
+  destinations:
+  - namespace: '*'
+    name: '*-spoke-*'
```
shapirov103: Two things here.

1. **Placement**: this file is not applied. `gitops/bootstrap/root-appset.yaml:23` syncs `{{ addonsRepoBasepath }}bootstrap` with `recurse: false`, and `config.yaml:13` sets `basepath: "gitops/"`, so only `gitops/bootstrap/` is read. On `main` this directory is applied by Terraform (`platform/infra/terraform/common/variables.tf:143`), and this PR removes Terraform, so it does not become live after the merge back either. If the projects are wanted, the file needs to move directly into `gitops/bootstrap/`.

2. **This line specifically**: `name: '*-spoke-*'` restricts the project to clusters whose names match that glob. Cluster names are customer-supplied and arbitrary, so a spoke named e.g. `prod-eu` would not match and its apps would be refused. `peeks-spoke-dev` / `peeks-spoke-prod` happen to match, so this looks correct on our clusters and fails on a consumer's. Suggest `name: '*'` (scoping via the resource whitelists instead), or dropping the file per option 1 in the summary.

allamand:
Thanks for the thorough review @shapirov103 — pushed `94ec0c76` addressing everything, plus two platform-bootstrap fixes we hit while validating Workshop Studio build `c9d02fda`.

### Your review

**Blocker — `spoke-workloads` AppProject never applied → chose Option 2 (keep the scoping):**
- Moved `appprojects.yaml` from `gitops/fleet/bootstrap/` (outside the path `root-appset.yaml` syncs, and no longer applied by Terraform after this merge) into `gitops/bootstrap/` so it is actually applied at the top level (`recurse: false`).
- Fixed the destination glob you flagged inline: `name: '*-spoke-*'` → `name: '*'`, since cluster names are customer-supplied and arbitrary; scoping stays via the resource whitelists. Skeletons keep `project: spoke-workloads`.

**Trivia:**
- Deleted the 0-byte `gitops/addons/clusters/peeks-hub/addons/backstage/values.yaml` (path no appset reads + baked-in cluster name).
- De-duped the `"✓ Keycloak configuration completed successfully!"` echo in `keycloak-config.yaml`.

**Terraform deletion (89 files):** confirmed intentional, left as-is — flagging that we treat the merge-back as a bootstrap-path cutover as you noted.

### Two extra fixes (platform bootstrap, found during validation)

- **Kargo ingress broke the shared `platform` ALB group.** It hardcoded an `HTTPS:443` listener + `ssl-redirect`, which triggers ACM cert auto-discovery-by-host and fails in CloudFront (`insecure`) mode → `buildModel` failed → keycloak/kargo lost their rules, backstage/argo lost targets. Kargo was the only ingress-bearing addon still on the raw upstream chart, so it couldn't switch on `insecure` like the others. Converted it to a local **umbrella chart** (`gitops/addons/charts/kargo`) wrapping the upstream chart as a dependency and rendering the API Ingress itself with the same `insecure` block-conditional as argo-workflows/backstage/keycloak (upstream ingress disabled). Mirrors the existing jupyterhub/karpenter umbrella pattern.
- **`idc:configure` diagnostics:** `keycloak_token()` surfaced a DB-down HTTP 500 as a cryptic `KeyError('access_token')`; it now raises a clear error distinguishing DB-down (500) from bad password (401).

Note: the `system-peeks` NodePool and `async-domain` seeding issues from that run needed no source change on this branch — the NodePool is already created by `platform-manifests-bootstrap` at sync-wave -1 (amd64) and `KEYCLOAK_DNS` already falls back to `config.local.yaml`; those failures were downstream symptoms of the aborted init. Re-validating on a fresh deploy now.

allamand:
While testing, some nodepools where not created for backstage on system-peeks custom nodepool. last commit should fix that, retesting now

gitops/addons/charts/kargo/values.yaml
```diff
@@ -0,0 +1,16 @@
...
+  # insecure=false: ALB serves HTTPS:443 with an HTTP->HTTPS redirect (real domain + cert).
+  insecure: "false"
+
+# Passed through to the upstream Kargo chart (dependency name `kargo`). The upstream
+# ingress is disabled here because this chart renders its own insecure-aware Ingress
+# (see templates/ingress.yaml).
+kargo:
+  api:
+    ingress:
+      enabled: false
```
shapirov103, line 16: Question on the path change here. This chart's ingress hardcodes `path: /kargo`, but the previous setup deliberately served Kargo at the host root. The workshop overlay `workshop/overlay/overlays/environments/control-plane/kargo/values.yaml` enables the ingress at `/` and its comment says "Kargo has no path prefix, so it serves at the host root (/)". Note the old registry entry already set `basePath: /kargo` with the same rewrite transform, and the overlay still overrode it back to root, so `basePath` alone didn't seem to be enough.

As I understand it Kargo doesn't handle being served under a path prefix. If that's right, then with `path: /kargo` the browser gets index.html fine, but the SPA then asks for its assets at the host root, which has no matching ALB rule, and the UI breaks. Worth explicitly checking the Kargo UI loads at `/kargo` with assets on the fresh deploy you're validating.

Two follow-ons depending on the answer. If `/kargo` does work, the workshop overlay is now dead config (it sets top-level `api.ingress.*`, which is both the wrong key path now that upstream values nest under `kargo:`, and aimed at the ingress this line disables) so it should be deleted, along with updating the Kargo section of `troubleshooting.md`. If it doesn't work, the ingress needs to go back to `/` and keep `group.order: '100'` so it doesn't shadow `/keycloak` and break SSO.

shapirov103 (commented):
Thanks, checked the fixes and they all look right: `appprojects.yaml` is now at `gitops/bootstrap/appprojects.yaml` so it lands inside the path `root-appset.yaml` syncs, both destinations are `name: '*'`, the 0-byte `peeks-hub` file is gone, and the keycloak echo is down to one. Blocker is closed as far as I'm concerned.

One question on the new Kargo chart, left inline.

gitops/addons/charts/kargo/values.yaml
allamand: Good catch — I validated this empirically on a fresh deploy and `/kargo` does work, so I went with your first follow-on (delete the dead overlay).

Findings on the fresh deploy (insecure/CloudFront hub):
- index.html emits `<base href="/kargo/">` because `kargo.api.basePath: /kargo` is set in the registry (`gitops/addons/registry/gitops.yaml`), and its assets are **relative** (`./assets/...`), so the browser resolves them under `/kargo/` — not the host root. So the SPA does not break.
- `GET /kargo/` = 200, `/kargo/assets/index-*.js` = 200, `/kargo/assets/index-*.css` = 200.
- `/kargo/healthz` (API via the `url-rewrite` transform that strips `/kargo`) = 200.
- `/keycloak` = 302 — not shadowed. Because `/kargo` is a specific prefix (not a `/` catch-all), it does not need `group.order`.

So the workshop overlay `workshop/overlay/overlays/environments/control-plane/kargo/values.yaml` was indeed dead config (top-level `api.ingress.*` — wrong key now that upstream values nest under `kargo:`, and aimed at the ingress this chart disables). Removed it and updated the Kargo note in `troubleshooting.md` in ac8cfc29.

allamand:
### E2E-test fixes applied to this branch (from MR !392 / `feat/pr-409` validation)

Following a full end-to-end test of the workshop on a fresh Workshop Studio event, the following fixes were committed here:

**Blockers**
- **FIX 1** — `applications/java/deployment/templates/kro/services.yaml`: Service names/selector (`java-app*`) did not match the `java-webservice` AppmodService, so the kro-generated Rollout stayed `Degraded` (`service "java-webservice" not found`, 0 pods) on the kro path. Renamed to `java-webservice`/`-preview`, matching the rust template. (KubeVela path unaffected — it creates its own Services.)
- **FIX 2** — `cluster-providers/kind-kro-ack`: the Keycloak `user_password` (= `USER1_PASSWORD`) was generated as a fresh random, while GitLab/IDE use `IDE_PASSWORD`, so `user1` could not log into GitLab with the documented `$USER1_PASSWORD`. Now seeds `USER_PASS=${IDE_PASSWORD:-…}` (mirrors kind-crossplane) → one user1 password everywhere.

**Priority 2/3**
- **FIX 3** — `gitops/addons/configs/... argo-workflows`: `argo-server` auth-mode was gated on `.Values.insecure`, so the workshop (insecure=true, TLS at CloudFront) ran `--auth-mode=client` and the UI's SSO login hit `GET /argo-workflows/oauth2/redirect` = **501**. The SSO block + keycloak client already exist, so switched to always `--auth-mode=sso` (like ArgoCD/Grafana/Backstage).
- **FIX 4** — `applications/java/kargo/project.yaml`: prod stage was `autoPromotionEnabled: true`, contradicting module 30.4 ("manual production promotion"). Set prod to manual. The previous auto-policy was a Kargo v1.7.5 webhook workaround; not needed on the shipped **v1.11.0** (verified: a server-side dry-run of a manual Promotion on prod is accepted).

**Region-agnostic (adapts #812 to this branch's ALB ingress)**
- `gitops/addons/charts/kro/resource-groups/manifests/appmod-service.yaml`: the RGD pinned the Ingress rule host to `*.us-west-2.elb.amazonaws.com`; outside us-west-2 the app ingress 404s. Made routing path-based (dropped host pinning, `ingress.host` default `""`). Because this RGD diverged from `main`, #812 would not land here cleanly at merge.

**Kargo `/kargo` subpath** (re: the earlier review thread): verified `/kargo` works end-to-end (assets + API + SSO), so the dead root-ingress overlay was removed. One residual **cosmetic** issue remains — the Kargo UI logo images use root-absolute paths and 404 under a subpath — filed upstream: **akuity/kargo#6849** (follow-up to akuity/kargo#6090).

Content-side fixes (broken LTC image, stale app count, DDB template copy step, Argo CD SSO wording) are in the content MR `feat/pr-409`.

Note: the `cicdpipeline` `Ready=False` / `"cluster mutated"` observation is **not introduced here** — the RGD is byte-identical to `main`; it stems from kro reconciling resources that other controllers continuously mutate (Argo Workflows/Events, the ECR-refresh docker secret). Cosmetic (the pipeline works); tracked as a broader kro concern.

gitops/addons/charts/kargo/values.yaml
allamand: only the image is not set see https://github.com/akuity/kargo/issues/6849

shapirov103: Nice! it is cosmetic, no issue.

shapirov103 (approved):
LGTM

elamaran11 (approved):
LGTM

PR #811, opened 2026-08-18
Replace CloudFront image delivery with in-service WebP catalog

5herlocked:
Rust microservice:
- Serve product images directly via Rocket's built-in FileServer instead of a hardcoded CloudFront/S3 URL. Real images live in static/images/ (checked into git as WebP for smaller transfer size); anything missing gets a small generated placeholder at startup so the catalog is never broken.
- image_url_for() prefixes the returned URL with APP_BASE_PATH (env var, empty by default) so external consumers get a correctly routable URL through this app's ingress path, while the pod itself only ever serves the unprefixed /product-images route -- nginx's existing rewrite-target strips the prefix before the request reaches the pod, exactly like every other route in this app (see platform-meta/templates/traits/path-based-ingress.yaml).
- Deployment templates (kubevela + kro) set APP_BASE_PATH=/rust-app to match their existing ingress path.
- Dockerfile and Dockerfile.arm copy static/ into the build stage and into the final image at the absolute path the relative!() macro expects.
- products.csv updated to reference the .webp filenames.

Next.js app:
- Product images are now served through the same shared ingress domain rather than an external CDN, so drop the stale CloudFront remotePatterns entry and mark product-image <Image> usages as unoptimized (next/image's server-side optimizer has no route to reach the rust service on its own).
- Fix a pre-existing Next.js 15 incompatibility: cookies() must be awaited before use (components/cart, components/wishlist).
- Regenerate pnpm-lock.yaml, which had drifted out of sync with package.json and was breaking --frozen-lockfile Docker builds.

Verified end to end with Finch: built both apps from their actual Dockerfiles, ran the Rust service against a local DynamoDB, and confirmed the containerized Next.js production server renders real WebP product images through the ingress-prefixed URL with zero proxying on the Next.js side.

*Issue #, if available:*

*Description of changes:*


shapirov103 (approved):
LGTM - great work!

PR #814, opened 2026-08-20
fix: scope S3Bucket IAMRoleSelector to instance-specific Bucket resource

### LINKED ISSUE #813: bug: S3Bucket RGD creates conflicting IAMRoleSelectors when multiple instances exist in same namespace
## Summary

When multiple `S3Bucket` Kro instances are created in the same namespace, each instance creates its own `IAMRoleSelector` scoped to the entire namespace and `s3.services.k8s.aws` group. This causes a **conflict**: two selectors with identical scope compete for the same ACK resources, causing the ACK S3 controller to enter an infinite reconcile loop (reading the resource every ~10 seconds) without ever creating the S3 bucket.

## Environment

- EKS Capabilities: ACK v46.137.1-eks-1, Kro v0.9.2-eks-3
- Affected RGD: `s3bucket.kro.run`
- File: `gitops/addons/charts/kro/resource-groups/manifests/rg-s3-bucket.yaml`

## Symptoms

- Second (and any subsequent) `S3Bucket` instance stays `IN_PROGRESS` forever
- The underlying ACK `Bucket` resource has **no status conditions at all** (empty `status: {}`)
- No `CreateBucket` call is ever made to the S3 API (confirmed via CloudTrail — zero events)
- The ACK controller polls the Bucket resource every ~10s indefinitely but never reconciles it
- Kro reports: `waiting for node "s3bucket": failed to evaluate readyWhen expression`

## Root Cause

The `IAMRoleSelector` template in the RGD uses a **namespace-wide scope**:

```yaml
spec:
  namespaceSelector:
    names:
      - ${schema.metadata.namespace}   # e.g. "default"
  resourceTypeSelector:
    - group: s3.services.k8s.aws
      kind: ""
      version: ""
```

This means **every instance** creates a selector that claims ALL `s3.services.k8s.aws` resources in the namespace. When a second instance is created:

- Instance 1 → creates `IAMRoleSelector` covering **all** S3 resources in `default`
- Instance 2 → creates a **second** `IAMRoleSelector` with the **same scope**

ACK resolves the first selector for all matching resources (both Bucket objects), and the second selector creates an ambiguous conflict. The second `Bucket` resource is then permanently stuck — the controller reads it every 10s but never writes a status or calls the AWS S3 API.

## Evidence (live cluster)

Two conflicting selectors observed on a real cluster deployment:

| Selector | Created | Status |
|---|---|---|
| `peeks-s3-kro-754909878517-84607729b8-s3` | first | ✅ ACK reconciled, bucket exists |
| `peeks-s3-kro-754909878517-ae34de8103-s3` | 6h later | ❌ ACK stuck in read loop |

- **16,642 GET calls** from the ACK capability role on the broken Bucket over 45 hours
- **Zero PATCH/UPDATE/status writes** on the broken Bucket in the same period (audit log confirmed)
- No `CreateBucket` CloudTrail event for the broken bucket

## Proposed Fix

**Option A — Scope the IAMRoleSelector to the specific resource instance (preferred)**

Add `kind`, `version`, and a `labelSelector` matching the Kro-owned label so each instance's selector only covers its own Bucket:

```yaml
  - id: iamroleselector
    template:
      apiVersion: services.k8s.aws/v1alpha1
      kind: IAMRoleSelector
      metadata:
        name: ${schema.metadata.name}-s3
      spec:
        arn: arn:aws:iam::${schema.spec.accountId}:role/peeks-cluster-mgmt-s3
        namespaceSelector:
          names:
            - ${schema.metadata.namespace}
        resourceTypeSelector:
          - group: s3.services.k8s.aws
            kind: Bucket
            version: v1alpha1
        labelSelector:
          matchLabels:
            kro.run/instance-name: ${schema.metadata.name}
```

> Note: Requires verifying ACK IAMRoleSelector supports `labelSelector` alongside `resourceTypeSelector`. If not, filling in `kind` and `version` at minimum reduces the blast radius.

**Option B — Single shared IAMRoleSelector as a platform-level resource (cleaner architecture)**

Remove the `IAMRoleSelector` from the `S3Bucket` RGD entirely. Deploy a single, shared `IAMRoleSelector` for `s3.services.k8s.aws` in each tenant namespace as a platform-managed resource (e.g., via the `platform-manifests` ArgoCD ApplicationSet). This mirrors the existing pattern used for `s3-peeks-spoke-staging`.

Benefits:
- No per-instance selectors → no conflicts regardless of how many S3Bucket instances are created
- Platform team manages IAM bindings centrally
- Backstage template no longer needs to know or pass the IAM role ARN

The `IAMRoleSelector` block would be removed from `rg-s3-bucket.yaml` entirely.

## Immediate Workaround (for existing broken deployments)

Delete the duplicate IAMRoleSelector created by the later instance. Since both selectors point to the same IAM role ARN, the first one already covers all resources:

```bash
kubectl delete iamroleselector <second-instance-name>-s3
```

> ⚠️ The selector has `kro.run/owned: "true"` — Kro will recreate it on the next reconcile. This workaround is only effective after the RGD is patched to use Option A or B above.

## Related Files

- `gitops/addons/charts/kro/resource-groups/manifests/rg-s3-bucket.yaml`
- `platform/backstage/templates/s3-bucket-ack-kro/skeleton/manifests/kro-s3-bucket.yaml`

allamand:
## Problem

Closes #813

When multiple `S3Bucket` Kro instances are deployed in the same namespace, each creates an `IAMRoleSelector` with identical scope — targeting **all** `s3.services.k8s.aws` resources in that namespace. This causes a conflict: the ACK S3 controller resolves the first (oldest) selector for every Bucket, and any subsequent Bucket enters an infinite read loop — polling the resource every ~10s indefinitely without ever calling `CreateBucket` in AWS.

**Observed impact on a live cluster:**
- 2nd `S3Bucket` instance stuck `IN_PROGRESS` for 45+ hours
- 16,642 `GET` calls from the ACK capability role, zero `PATCH`/`UPDATE` writes
- Zero `CreateBucket` events in CloudTrail
- ACK `Bucket` resource had empty `status: {}` throughout

## Fix

Add `resourceTypeSelector` (`kind: Bucket`, `version: v1alpha1`) and `resourceLabelSelector` matching `kro.run/instance-name` to scope each `IAMRoleSelector` exclusively to the Bucket resource owned by its Kro instance:

```yaml
# Before
resourceTypeSelector:
  - group: s3.services.k8s.aws
    kind: ""
    version: ""

# After
resourceTypeSelector:
  - group: s3.services.k8s.aws
    kind: Bucket
    version: v1alpha1
resourceLabelSelector:
  matchLabels:
    kro.run/instance-name: ${schema.metadata.name}
```

Kro automatically labels all child resources with `kro.run/instance-name`, and `resourceLabelSelector` is a supported field on the `IAMRoleSelector` CRD (`services.k8s.aws/v1alpha1`).

## Testing

Verified on a live EKS cluster (EKS Capabilities: ACK v46.137.1-eks-1, Kro v0.9.2-eks-3):

1. Applied the fix — Kro immediately reconciled both existing `IAMRoleSelector` resources with the new scoped spec
2. Touched the stuck `Bucket` resource to trigger ACK re-evaluation (needed once for pre-existing stuck instances; new instances work automatically)
3. ACK selected the correct scoped selector, assumed the IAM role, and created the S3 bucket within 30 seconds
4. Kro instance transitioned from `IN_PROGRESS` → `ACTIVE`; ArgoCD app from `Progressing` → `Healthy`

## Workaround for existing stuck instances

After deploying this fix, any pre-existing stuck `Bucket` resources need a one-time annotation touch to trigger ACK re-evaluation:

```bash
kubectl annotate bucket.s3.services.k8s.aws <bucket-name> \
  "services.k8s.aws/force-reconcile=$(date +%s)" --overwrite
```

### DIFF
--- a/gitops/addons/charts/kro/resource-groups/manifests/appmod-service.yaml
+++ b/gitops/addons/charts/kro/resource-groups/manifests/appmod-service.yaml
@@ -75,6 +75,10 @@ spec:
             - group: iam.services.k8s.aws
               kind: ""
               version: ""
+          # Scope to only ACK resources owned by this Kro instance (bug #813).
+          resourceLabelSelector:
+            matchLabels:
+              kro.run/instance-name: ${schema.metadata.name}
     - id: eksroleselector
       includeWhen:
         - ${schema.spec.components.dynamodb.enabled == true}
@@ -94,6 +98,9 @@ spec:
             - group: eks.services.k8s.aws
               kind: ""
               version: ""
+          resourceLabelSelector:
+            matchLabels:
+              kro.run/instance-name: ${schema.metadata.name}
     - id: dynamodbroleselector
       includeWhen:
         - ${schema.spec.components.dynamodb.enabled == true}
@@ -113,6 +120,9 @@ spec:
             - group: dynamodb.services.k8s.aws
               kind: ""
               version: ""
+          resourceLabelSelector:
+            matchLabels:
+              kro.run/instance-name: ${schema.metadata.name}
     - id: dynamodbtable
       includeWhen:
         - ${schema.spec.components.dynamodb.enabled == true}
--- a/gitops/addons/charts/kro/resource-groups/manifests/cicd-pipeline/cicd-pipeline.yaml
+++ b/gitops/addons/charts/kro/resource-groups/manifests/cicd-pipeline/cicd-pipeline.yaml
@@ -44,7 +44,9 @@ spec:
        apiVersion: services.k8s.aws/v1alpha1
        kind: IAMRoleSelector
        metadata:
-          name: ${schema.metadata.namespace}-iam
+          # Include instance name to avoid name collision when multiple CICDPipeline
+          # instances exist in the same namespace (bug #813).
+          name: ${schema.metadata.namespace}-${schema.metadata.name}-iam
        spec:
           arn: arn:aws:iam::${schema.spec.aws.accountId}:role/peeks-cluster-mgmt-iam
           namespaceSelector:
@@ -54,13 +56,17 @@ spec:
              - group: iam.services.k8s.aws
                kind: ""
                version: ""
+          # Scope to only the ACK resources owned by this Kro instance.
+          resourceLabelSelector:
+             matchLabels:
+                kro.run/instance-name: ${schema.metadata.name}
 
     - id: iamroleselectorEcr
       template:
        apiVersion: services.k8s.aws/v1alpha1
        kind: IAMRoleSelector
        metadata:
-          name: ${schema.metadata.namespace}-ecr
+          name: ${schema.metadata.namespace}-${schema.metadata.name}-ecr
        spec:
           arn: arn:aws:iam::${schema.spec.aws.accountId}:role/peeks-cluster-mgmt-ecr
           namespaceSelector:
@@ -70,13 +76,16 @@ spec:
              - group: ecr.services.k8s.aws
                kind: ""
                version: ""
+          resourceLabelSelector:
+             matchLabels:
+                kro.run/instance-name: ${schema.metadata.name}
 
     - id: iamroleselectorEks
       template:
        apiVersion: services.k8s.aws/v1alpha1
        kind: IAMRoleSelector
        metadata:
-          name: ${schema.metadata.namespace}-eks
+          name: ${schema.metadata.namespace}-${schema.metadata.name}-eks
        spec:
           arn: arn:aws:iam::${schema.spec.aws.accountId}:role/peeks-cluster-mgmt-eks
           namespaceSelector:
@@ -86,6 +95,9 @@ spec:
              - group: eks.services.k8s.aws
                kind: ""
                version: ""
+          resourceLabelSelector:
+             matchLabels:
+                kro.run/instance-name: ${schema.metadata.name}
 
     # ECR Repository for main application images
     - id: ecrmainrepo
--- a/gitops/addons/charts/kro/resource-groups/manifests/rg-s3-bucket.yaml
+++ b/gitops/addons/charts/kro/resource-groups/manifests/rg-s3-bucket.yaml
@@ -28,8 +28,14 @@ spec:
             - ${schema.metadata.namespace}
         resourceTypeSelector:
           - group: s3.services.k8s.aws
-            kind: ""
-            version: ""
+            kind: Bucket
+            version: v1alpha1
+        # Scope this selector to only the Bucket owned by this Kro instance.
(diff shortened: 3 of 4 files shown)


ROunofF (approved):
LGTM

elamaran11 (approved):
LGTM

allamand:
Extended the same scoping fix to the standalone **`DynamoDBTable`** RGD (`solutions/module1/rg-dynamodb-gen.yaml`), which had the identical namespace-wide `resourceTypeSelector` (`dynamodb.services.k8s.aws`, empty kind/version) with no `resourceLabelSelector` — two `DynamoDBTable` instances in one namespace would hit the same ACK read loop. Narrowed to `kind: Table/v1alpha1` + added `resourceLabelSelector: kro.run/instance-name`.

### Audit of all IAMRoleSelector usages
Reviewed every file defining `IAMRoleSelector`:
- ✅ `appmod-service.yaml`, `cicd-pipeline.yaml`, `rg-s3-bucket.yaml` — fixed in this PR
- ✅ `solutions/module1/rg-dynamodb-gen.yaml` — **added in this PR** (same bug)
- ⏭️ `multi-acct/templates/iam-role-selectors.yaml` and `platform-manifests/templates/ray-system-iamroleselectors.yaml` — **intentionally left as-is**: these are static, singleton, cluster-level grants (one selector per group per cluster namespace), not per-instance RGD selectors, so they don't hit the multi-instance conflict.

### Runtime validation (cluster, kro v0.9.2)
Confirmed the fix mechanism end-to-end on a live cluster:
- kro applies `kro.run/instance-name` (and `kro.run/owned: true`) to the ACK resources it creates → the `resourceLabelSelector` matches exactly.
- The unfixed RGD produces a namespace-wide `IAMRoleSelector` (`kind: ""`, no label selector), reproducing the #813 scope.

So `resourceLabelSelector: matchLabels: {kro.run/instance-name: <instance>}` correctly scopes each selector to its own instance's resources.

allamand:
Superseded by #822.

This PR was written against the pre-#529 layout and now conflicts. Part of the fix (the `appmod-service.yaml` `resourceLabelSelector` blocks) already landed via #529. #822 re-applies only the remaining pieces against the current `main`:
- `rg-s3-bucket.yaml`: scope to `kind: Bucket` + `resourceLabelSelector` (the core #813 fix)
- `workshop/solutions/module1/rg-dynamodb-gen.yaml`: same for `kind: Table` (file moved under `workshop/`)
- `cicd-pipeline.yaml`: instance-name in the IAMRoleSelector names (their labelSelector was already added by #529)

Leaving this open for review; can be closed once #822 is merged.

Issue #775, opened 2026-07-09
feat: hub self-adoption — pivot management from ephemeral kind to hub EKS


allamand:
## Problem

Currently the bootstrap kind cluster must remain **permanently alive** to host KRO/ACK controllers that manage `peeks-hub` and spoke clusters. This causes several issues:

- **Credential expiry**: The kind cluster uses EC2 instance role STS tokens (~1h) via a `aws-credentials` secret. After expiry, ACK controllers lose access → PIAs stuck, EksCluster stuck IN_PROGRESS, manual `task credentials:refresh` required.
- **Persistent EC2 dependency**: The kind cluster runs inside the IDE EC2. If the IDE is restarted or SSM session expires, the kind cluster is lost and the entire management plane is gone.
- **Non-idiomatic**: Using a local kind cluster as the long-running control plane for production EKS clusters is fragile.

## Proposed Solution: Hub Self-Adoption (CAPI Pivot Pattern)

Inspired by [Cluster API pivot](https://cluster-api.sigs.k8s.io/clusterctl/commands/move.html), the idea is to move the management of `peeks-hub` from the ephemeral kind cluster to the hub EKS cluster itself.

### Phases

**Phase 1 (today):** Kind bootstrap → creates peeks-hub
```
kind (ephemeral) ──KRO/ACK──→ creates peeks-hub EKS
                                         │
                                         ▼
                                  ArgoCD/KRO/ACK (EKS Capabilities)
```

**Phase 2 (proposed):** hub:pivot → peeks-hub self-manages
```
peeks-hub ──KRO/ACK (in-cluster)──→ adopts its own EksCluster resource
peeks-hub ──KRO/ACK──→ manages spoke clusters
kind → destroyed (no longer needed)
```

### Implementation Steps

1. **Install KRO + ACK on peeks-hub** (as EKS Capabilities — already done today)
2. **Export EksCluster manifest from kind** — serialize the current `EksCluster/peeks-hub` resource
3. **Apply on peeks-hub context** — with `adopt: true` or by setting the correct owner references so ACK adopts the existing AWS resources instead of recreating them
4. **Delete from kind** — remove the `EksCluster` resource from kind (ACK will not delete the AWS resources if adopted properly)
5. **Pivot spoke management** — move spoke `EksClusters` to peeks-hub context
6. **Destroy kind** — `kind delete cluster`

### Benefits

- ✅ No credential expiry (EKS PodIdentity, never temporary tokens)
[...]

Issue #647, opened 2026-05-15
RFC: GitOps diff preview on PRs — validate ArgoCD + kro + ACK changes before merge


allamand:
## Problem Statement

Multiple users have raised the need to **validate the full rendered output of GitOps changes before merging a PR**. Today, reviewing a PR that modifies Helm values, Kustomize overlays, kro ResourceGroupDefinitions, or ACK resource specs requires mentally parsing what ArgoCD will actually apply — which is error-prone and slows down reviews.

**User pain points:**
- "I changed a Helm value but I am not sure what manifests it affects across environments"
- "I modified a kro RGD template but cannot see what the downstream resources will look like"
- "ACK resource changes (S3, RDS, etc.) are hard to review without seeing the final spec"
- "We need an Atlantis-like experience for Kubernetes — plan before apply"

## Proposed Solution

Integrate [argocd-diff-preview](https://github.com/dag-andersen/argocd-diff-preview) (689⭐, presented at ArgoCon NA 2024 + EU 2026) into our CI/CD pipeline to automatically post rendered manifest diffs as PR comments.

### How it works

1. On PR open/update, a workflow clones base branch + PR branch
2. Spins up an ephemeral ArgoCD instance (via kind) or connects to a pre-configured one
3. ArgoCD renders all Application/ApplicationSet manifests from both branches
4. Generates a markdown diff and posts it as a PR comment

### What you see in the diff

| Layer | Visible in diff? | Notes |
|-------|-----------------|-------|
| Helm rendered manifests | ✅ Yes | Full YAML output after value substitution |
| Kustomize overlays | ✅ Yes | Merged patches visible |
| kro RGD definitions | ✅ Yes | Template changes visible |
| kro RGD instances (ResourceGroups) | ✅ Yes | Spec changes visible |
| ACK resource specs (S3Bucket, DBInstance) | ✅ Yes | Desired state changes |
| **kro-generated sub-resources** | ❌ No | Runtime controller output, not in Git |
| **ACK-provisioned AWS resources** | ❌ No | Cloud-side, not in Git |

### Extending to kro-generated resources (future)

To preview what kro will **actually produce** from a RGD change, we would need:
- The upcoming **kro CLI** (`kro render` / `kro plan`) to locally expand RGD templates into sub-resources
- A pipeline step that runs `kro render` on both branches and diffs the output
- This would give a complete "plan" including Deployments, Services, ConfigMaps that kro generates

**Proposed future flow:**
```
[...]

PR #818, opened 2026-08-24
feat(pod-identity): ack-pod-identity addon + ACK policy adoption in RGDs

allamand:
Two related changes around ACK resource adoption and addon pod identities.

Related issues: **#775** (hub self-adoption / CAPI pivot — the per-resource adoption semantics
established here answer its implementation step 3), **#813** (duplicate IAMRoleSelector scope
→ infinite ACK reconcile loop), **#647** (GitOps diff preview on PRs — natural home for the
drift test).

---

### Problem
KRO `HuggingFaceModel` instances stuck `IN_PROGRESS` on the `iamRole` node:
```
waiting for node "iamRole": readyWhen condition evaluated to false:
iamRole.status.conditions.exists(x, x.type == 'ACK.ResourceSynced' && x.status == "True")
```
The dependency `Policy` (ACK `iam.services.k8s.aws`) fails with **409 EntityAlreadyExists** on
`CreatePolicy` when the IAM policy already exists — e.g. a re-install into an account where a
previous run left the policy behind. ACK `Policy` is keyed by **ARN**, which the controller
cannot derive from `spec.name` alone, so `adopt-or-create` silently falls back to Create and
409s. The Role waits on the Policy and the whole instance stalls.

### Fix
Add `services.k8s.aws/adoption-fields` (templated ARN) alongside `adopt-or-create` on every
RGD `Policy` with a derivable ARN: `rg-huggingface-model.yaml` (s3-policy), `rg-eks.yaml`
(ack-iam, lbc, keycloak-config, cni-metrics-helper), `cicd-pipeline.yaml` (ecr-policy).
`pod-identity.yaml` gets `adopt-or-create` **only** (user-settable `spec.path`, no `accountId`
in the schema, so no reliable ARN — documented inline).

Scope: `Policy` resources only. Role/Table/Bucket/Repository are adopted by name.

---

### Problem
On `provider: kro-ack` clusters the `EksCluster` RGD created addon Pod Identity Associations
**unconditionally**, which gives three defects on a fresh install:

1. **Conflict with the Crossplane addon.** A pod identity association is keyed by
   `(cluster, namespace, serviceAccount)` and is unique in AWS, so the RGD and
   `crossplane-pod-identity` competed for the same key. Observed as
   `409 ResourceInUseException: The service account is already associated with a different IAM
   role`, leaving the app permanently `Degraded`.
2. **No per-addon conditionality.** The RGD cannot see the `enable_*` cluster labels, so it
   created identities for addons that are switched off.
3. **A wrong service account.** The RGD created the external-dns association for
   `external-dns/external-dns`, but the controller runs as `kube-system/external-dns-sa`. The
   association was useless and external-dns had no working identity at all — it fails with
   `no EC2 IMDS role found` on Route53 calls. Same wrong SA on hub and spokes.

### Fix
A new `gitops/addons/charts/ack-pod-identity` chart — the ACK mirror of
`crossplane-pod-identity`, rendering `iam.services.k8s.aws` Role/Policy and
`eks.services.k8s.aws` PodIdentityAssociation. One chart per backend follows the existing
crossplane/kro-ack convention (`abstractions/crossplane` vs `abstractions/kro`,
`clusters-crossplane.yaml` vs `clusters-kro.yaml`).

Registry selection is the **exact complement** of the existing entry, so a cluster gets
exactly one of the two and defect 1 cannot recur:

| App | selector |
|---|---|
| `pod_identities` (existing) | `provider NotIn ["kro-ack"]` |
| `ack_pod_identities` (new) | `provider In ["kro-ack"]` |

`Role.spec.policyRefs` / `PIA.spec.roleRef` are used instead of building ARN strings, so ACK
resolves the ARNs and orders the work.

### The scope split (the actual rule)
Because the association key is unique, the RGD and the chart must own **disjoint sets of
`(namespace, serviceAccount)`**. The 9 addon blocks (external-dns / lbc / keycloak-config ×
Policy/Role/PodIdentityAssociation) are commented out of `rg-eks.yaml` — commented rather than
deleted so the previous ownership stays visible.

| | owns |
|---|---|
| `EksCluster` RGD | cluster lifecycle (cluster/node roles, security groups, access entries, capability roles) + the genuinely chicken-and-egg bootstrap identities |
| `ack-pod-identity` chart | everything gated by an `enable_*` label |

Two identities deliberately stay with the RGD:
- `crossplane-system/provider-aws-{iam,eks}` — a provider cannot create the credentials it
  needs to run.
- `external-secrets/external-secrets-sa` — **structural, not cautious**: `external-secrets`
  syncs at sync-wave **-1**, earlier than any pod-identity app can run, and ESO must be
  authenticated before anything else can resolve its secrets. So its identity has to exist
  from cluster-creation time. The `eso` entry stays in the chart values at `enabled: false`
  (so the drift test keeps guarding its policy document) with an explicit warning not to
  enable it via overlay.

Four identities also stay in the RGD **for now**: `adot-collector`,
`cloudwatch-observability`, `cni-metrics-helper`, `kyverno-policy-reporter`. They map to real
registry addons (`observability.yaml`, `security.yaml`) or, for adot, to a scrape target in
the enabled `observability-aws` addon — they are merely disabled in the default config, and
`enable_*` is per-environment. Migrating them is blocked on verifying who actually owns
`amazon-cloudwatch/cloudwatch-agent` and `adot-collector-kubeprometheus`; doing it on an
assumed mapping would reproduce defect 3.

### CARM: `multi-acct` is the sole owner of IAMRoleSelector
ACK resolves which IAM role to assume per namespace **and** API group via `IAMRoleSelector`.
Without a selector covering a namespace, ACK creates the CRs and then **silently never
reconciles them** — no error, no events.

**INVARIANT: any cluster running this addon must be covered by `multi-acct`.**

`multi-acct` now **self-covers the cluster it runs on**, merging it into the effective map from
`global.clusterName` + a new `global.accountId` (the `aws_account_id` annotation). This is
required, not cosmetic: `.Values.clusters` is only ever populated with **spokes** (written by
`spokes:enable-*` into `configs/multi-acct/values.yaml` in the fleet-config overlay, and there
is no seed for that file in this repo), so nothing ever added the hub. Without self-coverage a
fresh install leaves the hub uncovered and every hub-side ACK resource fails silently — no pod
identity for lbc/external-dns/keycloak, i.e. no working ingress.

Derived from the annotation rather than a hardcoded name or the resource prefix, because
cluster names are customer-supplied. An explicit `.Values.clusters` entry still wins.

This chart briefly shipped its own selectors as a stopgap; that duplicated ownership, and two
selectors with an identical `(namespace, group)` scope put the ACK controller in an infinite
reconcile loop (empty `status{}`, zero AWS API calls — **#813**; identical role ARNs do not
help, the duplicated scope is the bug). The template was removed outright rather than gated
behind a flag.

Adding a cluster to `multi-acct` also gives it the `ec2` and `secretsmanager` selectors that an
on-hub `EksCluster` instance will need for **#775** — measured across the RGD's 39 active
resources: `eks` (18), `iam` (13), `ec2` (2, the ingress security groups), `secretsmanager` (2).

### Sync wave: -4, not 4
The crossplane twin sits at wave 4 because it must follow `crossplane-base` (wave 3), which
registers its CRDs. This chart has no such dependency — ACK is an EKS Capability, and CRD
readiness is gated by the chart's own PreSync wait hook. ArgoCD marks the app `Synced` when the
CRs are **applied**, not when `ACK.ResourceSynced=True`, so at wave 4 there was almost no slack
before the consumers (lbc / external-dns / ingress-class-alb at 5, keycloak at 6) — a pod can
start before its credentials exist and fail with `no EC2 IMDS role found` until the
`podidentity-pod-restart` CronJob rescues it. Lower bound is `multi-acct` at wave -5.

### Adoption annotations
Every resource carries `adopt-or-create` and `deletion-policy: retain`, so a re-install or a
re-created CR takes ownership of what already exists instead of failing, and a prune never
deletes a live AWS resource. The required inputs differ per resource, which is easy to get
wrong:

| Resource | Keyed by | `adoption-fields` needed? |
|---|---|---|
| Policy | ARN | **Yes** — otherwise create is attempted and 409s |
| Role | name | No — derived from `spec.name` |
| PodIdentityAssociation | clusterName + namespace + serviceAccount | No — and none is possible, `associationID` is AWS-generated |

The PodIdentityAssociation behaviour was established empirically: a CR with no
`adoption-fields` adopted a pre-created association and reported the **same** `associationID`,
with no duplicate in AWS.

### Drift guard
`values.yaml` was generated from `crossplane-pod-identity/values.yaml`, so the IAM policy
documents are identical by construction. `task test-pod-identity-drift` asserts
`policy.document` (semantic JSON), `policy.name`, `roleName`, `namespace` and `serviceAccount`
match for every shared identity. A divergent policy is a security bug (one provider granting
more than the other); a divergent namespace/SA breaks credentials *silently* — defect 3 above.

---

## Validation

Part 1: all four RGDs pass `kubectl apply --dry-run=server` against KRO.

Part 2, rendering:
- chart renders the expected set; all policy and assume-role documents parse as JSON;
  external-dns targets `kube-system/external-dns-sa`; PIAs carry no `adoption-fields` and
  Policies do.
- `appset-chart` against the registry produces `ack-pod-identities` with the correct hub
  destination and a selector that is the exact complement of `pod-identities`.
- `multi-acct` self-coverage rendered for four cases: empty map → cluster self-covered with 7
  groups + its namespace; map with 2 spokes → 3 namespaces × 7; cluster already in map → still
  7 (no duplicate, explicit account wins); no `accountId` → no resources.
- Drift test passes and **fails as intended on injected drift** (escalated `iam:*` statement,
  wrong service account).

Part 2, live on a hub + 2 spokes: all three `ack-pod-identities-*` apps `Synced/Healthy`, every
ACK resource `ACK.ResourceSynced=True`, 7 selector groups per cluster namespace with **zero**
duplicate-scope pairs, LBC and external-dns healthy with 0 auth errors, external-dns reporting
`All records are already up to date` (it was previously failing `no EC2 IMDS role found`), and
`/keycloak`, `/backstage`, `/argo-workflows` all returning 200.

## Gotchas worth knowing (propagation, not logic)

- **`addonsRepoRevision` is per cluster.** Each cluster secret carries its own, so pointing
  only the hub at a branch leaves the spokes rendering from the old one — the new chart path
  then does not exist and the app fails with `app path does not exist`.
- **Registry changes need `cluster-addons` refreshed.** A `valuesObject` addition (e.g.
  `global.accountId`) reaches an app only after the `cluster-addons` **Application** re-renders
  the appset-chart. Refreshing the generated Application, or even the addon ApplicationSet, is
  not enough — until then the app runs with stale values and can compute an empty desired
  state.
- **ArgoCD can sync from a cached copy of the overlay repo**, so pushing an overlay and
  flipping a cluster label in quick succession may apply the pre-overlay values. Confirm
  ArgoCD has picked the overlay up first.

## Note for existing environments only

None of the above requires a migration on a fresh deployment: kind applies the RGD with the
addon blocks already commented out, so the chart creates those identities from the start —
nothing to adopt and no role to repoint.

Existing clusters where the RGD already created these identities do need an ordered handover
(`deletion-policy: retain` must be applied and observed **before** the blocks are removed,
otherwise ACK deletes the live roles and associations). That runbook, with the reasoning and
the failure modes, is in `docs/platform/ack-pod-identity-design.md` and is out of scope for
this PR.

## TODO / follow-ups

- [ ] **Add `task test-pod-identity-drift` to CI.** This repo has no CI workflow other than
      `renovate.yml`, so the guard only runs if someone remembers — a false sense of safety for
      a check whose purpose is catching a silent security regression. Should run on PRs
      touching `gitops/addons/charts/{crossplane,ack}-pod-identity/**`; see **#647**.
- [ ] Write `provider` into whatever **seeds** the cluster config (`kro-clusters` values
      template / RGD schema) rather than only the live secret, or a re-seed silently drops it —
      the same failure mode as `overlay_repo_url`.
- [ ] Migrate the four currently-disabled identities once their real namespace/serviceAccount
      owner is verified.
- [ ] Remove the now-dead `external-dns/external-dns` association and the superseded
      `*-lbc-role` / `*-keycloak-config-role` / `*-external-dns-role` roles. Careful:
      `*-external-dns-role` carries the `*-ack-iam` policy, whose name suggests it may be
      referenced elsewhere.


shapirov103:
@allamand I see commits still coming, please indicate when ready for review.

allamand:
@shapirov103 I've validated this PR end to end and it was ok

shapirov103 (approved):
LGTM

PR #823, opened 2026-08-31
fix(hack): don't share zsh history between parallel terminals

allamand:
## Change
`hack/.zshrc` set `no_share_history` / `unsetopt share_history` **before** `source $ZSH/oh-my-zsh.sh`. oh-my-zsh's `lib/history.zsh` runs during that source and forces `setopt share_history`, silently overriding the earlier setting — so parallel terminals ended up sharing live history anyway.

Move the history-sharing override to **after** the oh-my-zsh source so it actually takes effect:

- `unsetopt share_history` — each session keeps its own live history (commands from one terminal are not injected into other already-running sessions).
- `setopt inc_append_history` — commands are still appended to `HISTFILE` as they run, so history is persisted across sessions.

A comment is added at the original (pre-source) location pointing to where the real override now lives, to prevent this from being reintroduced.

## Scope
Touches only `hack/.zshrc`. No prompt/theme changes.

> Note: an earlier revision of this PR also changed `hack/.p10k.zsh` separators; that part has been reverted and is no longer included.

## Testing
- Opened multiple parallel terminals in the workshop IDE (Amazon Linux 2023, code-editor-server) and confirmed each session keeps independent live history while `HISTFILE` still accumulates commands across sessions.

### DIFF
--- a/hack/.p10k.zsh
+++ b/hack/.p10k.zsh
@@ -168,24 +168,24 @@
   fi
 
   # Separator between same-color segments on the left.
-  typeset -g POWERLEVEL9K_LEFT_SUBSEGMENT_SEPARATOR='\u2502'
+  typeset -g POWERLEVEL9K_LEFT_SUBSEGMENT_SEPARATOR='\uE0B1'
   # Separator between same-color segments on the right.
-  typeset -g POWERLEVEL9K_RIGHT_SUBSEGMENT_SEPARATOR='\u2502'
+  typeset -g POWERLEVEL9K_RIGHT_SUBSEGMENT_SEPARATOR='\uE0B3'
   # Separator between different-color segments on the left.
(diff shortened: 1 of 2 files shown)


hmuthusamy (approved):
LGTM

hmuthusamy (approved):

shapirov103 (approved):
LGTM

shapirov103 (commented):
why do we care about history sharing? Is there any specific tangible issue that arose because of it? 

I can see how history sharing can be beneficial

allamand:
History will be share but not in live, I found it problematic when live history in different windows is shuffle at runtime, you may find hart to just execute your last command in a window because you work longer in a second one

shapirov103:
> History will be share but not in live, I found it problematic when live history in different windows is shuffle at runtime, you may find hart to just execute your last command in a window because you work longer in a second one

Gotcha

shapirov103 (approved):
LGTM

PR #824, opened 2026-08-31
fix(scripts): restore argocd-utils.sh and source it in recover-argocd…

allamand:
This is used by participants to manually force resync through the shell command : argocd-sync

recover-argocd-apps.sh calls 12 helper functions (get_infrastructure_report, detect_sync_timeout_pattern, terminate_argocd_operation, sync_argocd_app, verify_keycloak_secrets, refresh_keycloak_dependent_apps, verify_namespace_exists, verify_kubevela_dependencies, recover_stuck_workflows, detect_null_phase_workflows, trigger_workflow_manually, generate_dependency_report) but only sourced colors.sh.

These functions live in argocd-utils.sh, which was lost when scripts/ was migrated out of platform/infra/terraform/scripts/. As a result every recovery phase silently no-ops: each missing function prints "command not found" on stderr, but the script has no `set -e`, so it continues and still reports "Recovery complete". The early `exit 0` (all apps healthy) hid this in the common case — the bug only surfaces when apps are actually stuck.

- Restore scripts/argocd-utils.sh (complete 12-function version from 13b67f82).
- Source it in recover-argocd-apps.sh right after colors.sh.

Verified: bash -n OK on both; sourcing resolves all 12 functions; running the recovery path with a stubbed kubectl (simulated stuck app) drops "command not found" occurrences from 6 to 0.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/scripts/argocd-utils.sh
+++ b/scripts/argocd-utils.sh
@@ -1,0 +1,1534 @@
+#!/bin/bash
+
+# Robust ArgoCD utility functions
+
+export CORE_APPS=(
+    "external-secrets"
+    "argocd"
+    "gitlab"
+)
+
+export BOOTSTRAP_APPS=(
+    "bootstrap"
+    "cluster-addons"
+    "clusters"
+    "fleet-secrets"
+)
+
+# Infrastructure Verification Functions
+
+# Verify cluster infrastructure health
+verify_cluster_infrastructure() {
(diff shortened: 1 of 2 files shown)


shapirov103 (approved):
LGTM

elamaran11 (approved):

PR #31, opened 2024-09-26
Adding Argo-Rollouts-Metrics-Driven-Deployment.md

Eli1123:
This adds to the previous metrics driven deployment, and shows how to add metrics driven progressive delivery. Has some overlap since it built on the previous step I kept that for extra clarity. 

Handing to Mikhail to review.

*Issue #, if available:*

*Description of changes:*
Added the doc Argo-Rollouts-Metrics-Driven-Deployment.md and edited Argo-Rollouts-Metrics-Driven-Deployment.md for minor changes.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #825, opened 2026-08-31
fix(validation): make backstage-auth.sh zsh-safe (guard bash-only RETURN trap)

allamand:
## Problem
`scripts/validation/backstage-auth.sh` registers a cleanup with `trap ... RETURN` (bash-only). When the helper is **sourced in the workshop IDE's default zsh shell**, zsh errors with `trap: undefined signal: RETURN` and — under `set -euo pipefail` — aborts `backstage_get_token`, so `BS_TOKEN` is never set. Automated/workshop validation then has to wrap every call in `bash -lc`.

## Fix
Register the RETURN cleanup trap **only under bash** (`[ -n "${BASH_VERSION:-}" ]`). Under zsh the function now completes and sets `BS_TOKEN`; the temp files live under /tmp and are harmless if left uncleaned. No behavior change under bash.

Validated: `bash -n` clean, and sourcing under zsh no longer raises the RETURN trap error.

## Context
Fixes the secondary finding reported in platform-engineering-on-eks issue #31 (Rust CD non-regression testing).

### DIFF
--- a/scripts/validation/backstage-auth.sh
+++ b/scripts/validation/backstage-auth.sh
@@ -12,7 +12,13 @@ set -euo pipefail
 backstage_get_token() {
   local COOKIE_JAR
   COOKIE_JAR=$(mktemp)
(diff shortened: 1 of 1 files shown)


shapirov103 (approved):
LGTM

hmuthusamy (approved):

PR #852, opened 2026-09-02
fix(keycloak): make split-brain detector read live cluster state from…

allamand:
… jgroups_ping

The detector never worked: get_cluster_members() ran `kubectl exec <pod> -- kubectl logs`, i.e. kubectl INSIDE the keycloak container, which has no kubectl binary, so it always failed and fell back to "0". It therefore always logged 'cluster size 0', never matched the size==1 condition, and never healed. Even fixing the exec bug, the approach was unreliable: it grepped the boot-time 'Received new cluster view (N)' log line, which ages out of the log buffer, so after a while there is nothing to grep.

Rewrite it to read the authoritative, always-current source: the JGroups JDBC_PING discovery table (jgroups_ping) in the DB Keycloak actually uses at runtime (postgres). Detection:
  - >1 coordinator (coord=true) => split-brain (members partitioned)
  - a registered member IP with no matching live pod => stale/isolated member Heal conservatively by restarting only the NON-coordinator pod(s) so they rejoin the coordinator's view (preserves the primary view, avoids a full outage).

Runs on the postgres:17.4-alpine image (has psql; already used by this chart), adds curl+jq at start, and performs the pod list/delete via the Kubernetes API using the mounted ServiceAccount token (image has no kubectl). RBAC no longer needs pods/exec.

Verified live: correctly reports 'Ready pods: 2 | members: 2 | coordinators: 1 => healthy' against the running cluster (the old script reported 0).

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/gitops/addons/charts/keycloak/templates/split-brain-detector.yaml
+++ b/gitops/addons/charts/keycloak/templates/split-brain-detector.yaml
@@ -5,65 +5,113 @@ metadata:
   namespace: keycloak
 data:
   detector.sh: |
-    #!/bin/bash
-    set -euo pipefail
+    #!/bin/sh
+    set -eu
 
     NAMESPACE="keycloak"
-    
-    log() {
-        echo "[$(date +'%Y-%m-%d %H:%M:%S')] $*"
+    # JGroups cluster name used by Keycloak's Infinispan channel (see logs: "channel `ISPN`").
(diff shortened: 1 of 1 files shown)


shapirov103:
I was about to ask - if more than one replica jgroups discovery makes sense, why closed?

allamand:
The discovery was not accurate I closed this one but addressed this in a separate one: fix the split brain code, add pod-anti affinity..

PR #835, opened 2026-09-01
Release v0.3.0-rc3 — ingress, teardown, pod identity, analysis gates

allamand:
## Release v0.3.0-rc3

Integration of the v0.3.0-rc3 workshop fixes into `main`. Aggregates PRs #827, #828, #829, #830, #831, #832, #834 (sub-issues of #584). Validated read-only against the kro-c1 hub; the graceful-teardown path (#830) is not yet E2E-tested (a full provision→destroy on a reused account is the remaining validation for #34/#35/#36).

### Changes

**Ingress path routing 404 (#832)**
- Remove the bogus `domain: "*.elb.us-west-2.amazonaws.com"` host-header from the KubeVela `path-based-ingress` traits — it rendered an ALB host-header condition that never matched the real ALB DNS, so `/progressive`, `/northwind-app` etc. 404'd. Set `domain: ""` (the pattern rust/java already use).
- Files: dotnet `northwind`, java `progressive`, and 2 rust `platform-meta` examples.

**Progressive-delivery analysis gates (#828 — issues #42/#43/#44)**
- Functional gate targets the canary `${schema.metadata.name}-preview` service via CEL (Argo Rollouts does not substitute `{{args}}` in `job` providers).
- Performance gate runs a **real** Apache Bench load test (`httpd:alpine`, which ships `ab`) against the preview service and **rolls back** when `Time per request` exceeds the `performanceGate.extraArgs` threshold.
- Metrics gate query uses the AMP scraper label names: `k8s_container_name`→`container_name`, `k8s_namespace_name`→`namespace`.

**Spoke observability / Argo Rollouts AMP auth (#831 — issues #41/#42)**
- Argo Rollouts IRSA role + spoke OIDC provider registration so SigV4 AMP queries authenticate on spoke clusters.
- Re-seed `observability` after spokes are ACTIVE so spoke AMP scraper values reach Secrets Manager / the ArgoCD cluster secret (fixes hub-only scraper, #41).
- `ack-pod-identity` / `crossplane-pod-identity` charts + `pod-identities` overlays.

**Graceful teardown + Crossplane deletion ordering (#830 — issues #34/#35/#36/#593)**
- `hub:destroy-addons` rewritten for both providers: dynamic ApplicationSet discovery (kubectl, not static registry) + ownerReference child-app detection + skip-empty waits; ordered workloads→spokes→addons→infra→AWS-sweep teardown.
[...]

PR #860, opened 2026-09-03
fix(install): restart ACK controllers after hub EKS ACTIVE to inject Pod Identity creds

allamand:
## Problem

On fresh accounts, `task install` fails at `hub:wait-for-eks` after ~40 min with:
```
cniMetricsHelperPolicy.status.ackResourceMetadata.arn: no such key: arn (data pending)
```

**Root cause**: The install DAG calls `hub:claim` before ArgoCD is deployed on the hub. When KRO starts reconciling the `eksclusterwithvpc` RGD, it immediately tries to create ACK IAM resources (Policy, Role, PodIdentityAssociation). But the ACK IAM controller has no AWS credentials yet — its Pod Identity Associations are created by the `ack-pod-identities` ArgoCD AppSet (wave -4), which only deploys once the ArgoCD EKS Capability is up on the hub. Since `hub:seed` (which deploys ArgoCD) runs *after* `hub:wait-for-eks`, the ACK controllers start credential-less and KRO gets stuck.

## Fix

Added two new tasks between `hub:claim` and `hub:wait-for-eks`:

1. **`hub:wait-for-hub-active`** — waits for hub EKS to be ACTIVE in AWS (gates on hub being up so ArgoCD + ack-pod-identities AppSet have deployed)
2. **`hub:restart-ack-controllers`** — waits for ACK PIAs to appear, then force-restarts ACK controller pods so the EKS Pod Identity webhook injects fresh AWS credentials before KRO resumes reconciliation

## Test plan
- [ ] Fresh account deploy with `task install` should no longer fail at `hub:wait-for-eks`
- [ ] Existing account re-run should be unaffected (tasks are best-effort / idempotent)

## Related
- PR #835 (v0.3.0-rc3 teardown validation) — install failure observed during teardown testing on kro-c1

### DIFF
--- a/cluster-providers/kind-kro-ack/Taskfile.yaml
+++ b/cluster-providers/kind-kro-ack/Taskfile.yaml
@@ -153,6 +153,17 @@ tasks:
       - task: kro:apply-rgds
       - task: hub:wait-for-domain
       - task: hub:claim
+      # After hub:claim submits the eksclusterwithvpc, the KRO RGD immediately tries to
+      # reconcile ACK IAM/EKS resources (cniMetricsHelperPolicy etc.). Those resources
+      # need ACK controllers to have AWS credentials, injected via EKS Pod Identity.
+      # The PIAs are created by the ack-pod-identities ArgoCD AppSet (wave -4), which
+      # deploys once ArgoCD is up on the hub. However, the EKS Capabilities install ACK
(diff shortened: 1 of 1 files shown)


allamand:
Closing — incorrect analysis. ACK in kind-kro-ack is an EKS Capability, not pods. Will reopen with correct fix.

Issue #877, opened 2026-09-07
Teardown: CFN delete leaves orphaned spoke clusters + hub SGs block IDE VPC deletion (DELETE_FAILED)


allamand:
## Summary

On teardown of a Workshop Studio event (CFN stack delete), the **spoke EKS clusters are left orphaned** and **leftover hub security groups block the IDE/hub VPC deletion**, leaving the CloudFormation stack in `DELETE_FAILED` and incurring ongoing cost + reused-account collisions.

Observed empirically tearing down a `kind-kro-ack` event (hub + `peeks-spoke-dev` + `peeks-spoke-prod`, hub sharing the IDE VPC).

## What happens

CFN stack delete triggers the buildspec Delete branch which runs, via SSM on the IDE:

```
cd .../platform-on-eks-workshop/workshop && task destroy
```

Result after the stack finishes:
- ✅ IDE resources + hub cluster deleted
- ❌ **`peeks-spoke-dev` and `peeks-spoke-prod` still `ACTIVE`** (+ their VPCs, node roles)
- ❌ Stack ends in **`DELETE_FAILED`**: `The vpc 'vpc-…' has dependencies and cannot be deleted` — caused by leftover hub SGs in the IDE VPC (`eks-cluster-sg-peeks-hub-*`, `k8s-traffic-*`, `k8s-platform-*`, `peeks-hub-ingress-http/https`, `peeks-hub-platform-alb-sg`, `peeks-amg-sg`, `rds-mysql-sg-*`).

## Root cause (the ordering is NOT the problem)

`workshop/Taskfile.yaml`'s `destroy` task **already** implements the correct order — *"drain workloads from spokes → delete spokes → destroy hub"*. So this is **not** a missing teardown phase. The two real defects are:

1. **The CFN Delete path does not wait for `task destroy` to complete.** A full `task destroy` (spoke EKS cluster deletion) takes ~30–40 min, but the `ClustersStackDeploy` Delete CodeBuild finished in ~12 min and reported `SUCCEEDED`. The long spoke-deletion phase was cut off → spokes orphaned. The SSM `task destroy` needs to be polled to genuine completion (with a sufficient timeout), and its real exit status honoured.

[...]

PR #887, opened 2026-09-09
release: v0.3.0-rc4 → main (teardown robustness + spoke provisioning fixes)

allamand:
## Release / integration PR — `release/v0.3.0-rc4` → `main`

Umbrella PR carrying the accumulated **v0.3.0-rc4** work on appmod-blueprints into `main`. Mirrors the platform-engineering-on-eks side (GitLab !577).

The main theme is **reused-account teardown robustness** + fixing two **spoke-provisioning** regressions surfaced along the way.

### Teardown robustness (the sweep is the backstop when `task destroy` is cut short / partial)
- **#878** — sweep: reap orphaned spoke clusters + clear leftover hub security groups in the IDE VPC (#877). Leftover hub SGs (via circular refs) and orphaned spokes were blocking CloudFormation from deleting the IDE VPC.
- **#881** — sweep §6b: retry `delete_cluster` until the ACK capabilities finish deleting (~25 min). A single fixed short wait raced ACK (~15-20 min) and left spokes orphaned.
- **#882** — sweep: **authoritative final re-sweep of recreatable resources + ENI gate**. Sections that delete CloudFront/ALB/RDS/AMP run *before* the hub cluster is deleted; while the hub is alive its ACK/Crossplane/addon controllers re-create them → orphaned once the hub is gone. The new final pass runs *after* the hub and spokes are deleted (controllers gone) and waits for `cloudfront_managed` / `amp_collector` / `RDSNetworkInterface` ENIs to detach so CFN can delete the subnets.

### Spoke provisioning fixes
- **#885** — declare `managementRegion` in the **`EksclusterWithVpc`** RGD schema. It was declared on the `EksCluster` RGD (hub) and set on the instances, but not on the `EksclusterWithVpc` RGD the spokes use → kro rejected the field (`field not declared in schema`) → spokes never created.
- **#886** — never derive the GitLab/fleet-config domain from the **platform** CloudFront. The `GIT_DOMAIN` fallback used `{{.DOMAIN}}` (the `<prefix>-hub-platform` distribution) for the GitLab URL; when `private/git-domain` wasn't written yet, `set-overlay-repo` wired `overlay_repo_url` to the wrong host → the `clusters-kro` ApplicationSet pointed at a dead host → no spoke apps generated.

### Other fixes
- **#883** — `kind-kro-ack`: set `managementRegion` to the deploy region (fixes non-us-west-2 deploys).
- **#884** — Ray bucket: omit S3 `LocationConstraint` in `us-east-1`.
- **#880** — kubevela functional gate emits `Testing for`/PASS/FAIL like the kro path.
- **#879** — remove the obsolete `fix-bootstrap-post-clone.sh` recovery script.

### End-to-end validation (with the platform-side !581 + !583 teardown permissions)
Fresh deploy + teardown on a **reused account**:
- Deploy → **CREATE_COMPLETE**, all 3 clusters `ACTIVE` (hub + `spoke-dev` + `spoke-prod`); `clusters-kro` ApplicationSet generates cleanly.
- Teardown → **DELETE_COMPLETE, 0 `AccessDenied`, no manual intervention**: CloudFront distribution + VPC origin, clusters/spokes, and all hub/spoke secrets deleted; no orphaned ENIs blocking the VPC.

> Note: the workshop clones appmod-blueprints at runtime, so these ship via the rc4 default branch; the companion CFN/IAM changes (SharedRole teardown permissions, `secretsmanager:DeleteSecret`) are in platform-engineering-on-eks !581 / !583.


csantanapr (approved):
LGTM

workshop/scripts/sweep-workshop-resources.py
shapirov103, line 1: why is it under scripts? should not this be in the workshop directory?

workshop/scripts/sweep-workshop-resources.py
```diff
@@ -40,6 +40,53 @@ def log(msg):
     print(f"[sweep] {msg}", flush=True)
 
 
+hub_vpc_id = None
+spokes = [f"{prefix}-spoke-dev", f"{prefix}-spoke-prod"]
+
+
+def _revoke_sg_rules(sg):
```
shapirov103, line 47: How where these SG rules provisioned in the first place? Unclear why we need a script to clean this up. Should we focus on proper lifecycle management instead - depending on how this was provisioned.

shapirov103 (commented):
Please see my questions.

workshop/scripts/sweep-workshop-resources.py
allamand: yes true

workshop/scripts/sweep-workshop-resources.py
allamand: yes maybe, the sweep is kind of workarround in case things did not clean well. can be lots of reason, a rds launch somewhere, another thing launch also, we want the solution to be able to clean, but it is very difficult to properly fix things at delete time as they can come from many different places. but In fine when you are in delete script, at this time it is either fail delete or try to force it to succeed. That does not mean that we shouldn't fix proper things prior that, but again, not easy to do

allamand:
Moved to `workshop/scripts/sweep-workshop-resources.py` in PR [#889](https://github.com/aws-samples/appmod-blueprints/pull/889) + companion platform MR updates the buildspec reference.

workshop/scripts/sweep-workshop-resources.py
shapirov103: Got it, stopgap for now, let's plan to log issues if we see standard teardown leaves behind. I do recognize we have orphaned resources hanging in the account, affecting repeatability.

shapirov103 (approved):
LGTM

Issue #836, opened 2026-09-01
Tech debt: provider orchestration is shell-in-YAML, with silent-failure traps


shapirov103:
## Problem

The cluster-provider Taskfiles have grown into large programs written as shell embedded in YAML. `cluster-providers/kind-kro-ack/Taskfile.yaml` is ~1,900 lines with 53 tasks; `kind-crossplane/Taskfile.yaml` is ~1,600 with 45. Much of that is multi-line `bash` inside `cmds:` blocks doing control flow, polling loops, JSON assembly and AWS orchestration.

This is hard to read, effectively untestable, and it fails in ways that are specific to the embedding rather than to the logic. Three concrete instances found while implementing the `domainResolver` contract (PR on `feat/cluster-provider-parity-cloudfront`):

**1. `errexit` + `&&` silently aborts with no diagnostics.** These Taskfiles set `set: [errexit, nounset, pipefail]`. Under go-task's mvdan/sh interpreter, a failing `[ ... ] && cmd` as the last statement of a block makes the enclosing compound return non-zero and aborts the script. Observed: a `domain:resolve` implementation exited 1 *before printing any of its six diagnostic lines*, so the operator saw only `task: Failed to run task ...: exit status 1`. The same script ran correctly under `bash -e`. Debugging required tracing with `set -x` inside the task.

**2. Background jobs are killed.** `( ... ) &` inside a `cmds:` block is terminated when the task body exits — verified that the redirect target was never even created. `install:phase1-observability-bg` therefore never ran its 30+ minute AMP seed on either provider; only a fast-failing child ever produced log output, which is why it looked like it worked. Fixed with a double-fork (`sh -c '... &'`), which is not obvious and needs a comment explaining it.

**3. Template-time vs run-time variable evaluation is invisible.** go-task evaluates global `vars:` with `sh:` once at parse time. A task that updates the underlying config cannot change `{{.VAR}}` for the rest of the process, so consumers silently use stale values. Working around this required running a phase as a nested `task` process purely to force a re-parse.

None of these are logic bugs. They are all artifacts of writing a program in shell-inside-YAML, and each one failed *silently*, which is the worst property for provisioning code.

## Additional smells

- `yq '.domain'` returns the literal string `"null"` for a missing key. Non-empty, so it defeats `[ -z ... ]` guards; an `ingress_domain_name: "null"` could reach the cluster secret. Every config read needs `// ""` plus a `"null"` check by hand.
[...]

Issue #833, opened 2026-09-01
kind-crossplane cannot provision hub into an existing VPC (provider capability parity gap)


shapirov103:
## Summary

`kind-crossplane` cannot provision the hub into a **pre-existing (customer-supplied) VPC**. `kind-kro-ack` can. This is a capability gap between two cluster providers that are supposed to be interchangeable.

A customer running in their own account, who already has a VPC governed by their network team, currently cannot use the `kind-crossplane` provider at all. Raised by the `open-agentic-platform` consumer.

## Requirement: providers must be at capability parity

**Every cluster provider must expose an identical capability set. Adding a capability to one provider and not the other is not acceptable.** `cluster-providers/README.md` defines the providers as pluggable behind one contract, and `config.yaml` is a single shared schema. When a capability lands in one provider only:

- `config.local.yaml` silently means different things depending on `clusterProvider`. `hub.network.vpcId` is honoured by `kind-kro-ack` and ignored by `kind-crossplane`, with no error.
- Consumers cannot switch providers without re-validating, which defeats the pluggable design.
- The gap is invisible until provisioning either fails or, worse, silently creates a second VPC.

Any PR adding a provider capability should either implement it for all providers or open a tracking issue like this one **and** make the unimplemented provider fail fast with a clear message rather than ignore the input.

Immediate follow-up: `kind-crossplane` should reject `hub.network.vpcId` with an explicit "not supported by this provider, see #<this issue>" error instead of silently ignoring it and creating a VPC.

## Current state (verified)

| | existing-VPC import | mechanism |
|---|---|---|
| `kind-kro-ack` | supported | `HUB_VPC_ID`/`HUB_SUBNET_1`/`HUB_SUBNET_2` vars (`Taskfile.yaml:32-37`); `hub:claim` selects kind `EksCluster` + `network` block when `vpcId` set, else `EksclusterWithVpc` + `cidr` block |
| `kind-crossplane` | **missing** | only `hub.vpcCidr` (`Taskfile.yaml:49`, used at `:426`); always creates a VPC; VPC id derived post-hoc from the finished cluster (`:647-654`) |

kro-ack gets this cheaply because KRO exposes two RGD kinds and the Taskfile picks one.

## Why this is not a small change on crossplane

[...]

PR #888, opened 2026-09-10
Reserve the CloudFront hostname up front, not mid-install

shapirov103:
## What this does

Lets a consumer install the platform and reach it over HTTPS **without owning a domain**, using a free `*.cloudfront.net` hostname. It also removes the asynchronous domain machinery that previously existed for this case, because a change of ordering makes it unnecessary.

`kind-crossplane` and `kind-kro-ack` both work with this. No pre-existing VPC is required.

## The problem was ordering, not domains

The platform needs its ingress hostname **at install time**: Keycloak realm URLs, the OIDC issuer, every ingress host, and the ArgoCD and Backstage base URLs are derived from `domain`, and ArgoCD starts deploying addons at the end of the install.

With no registered domain the obvious approach is circular:

```
platform needs a hostname AT install time
  → hostname comes from a CloudFront distribution
    → the distribution needs an origin
      → the origin is the platform ALB
        → which is created BY the install that needs the hostname
```

Everything downstream came from working around that cycle: the ALB had to be pre-created, so the VPC had to pre-exist, so `kind-crossplane` could not offer CloudFront at all (#833), and the hostname only arrived mid-install, so it had to be polled for and passed through a file.

## The fix

`create-distribution` returns `DomainName` immediately and does **not** validate the origin at creation time. So a distribution can be created with a deliberately unresolvable placeholder origin before any infrastructure exists, and `domain` becomes an ordinary static config value.

```bash
domain=$(scripts/cloudfront-reserve-domain.sh)   # ~1s: no VPC, no ALB, no waiting
# write domain + insecure: true into config.local.yaml
task install                                     # unchanged, knows nothing about CloudFront
scripts/cloudfront-attach-origin.sh              # swap the placeholder for a VPC origin
```

Two shipped, reusable scripts, so an external consumer needs no CloudFront code of its own:

- **`scripts/cloudfront-reserve-domain.sh`** — pre-install. Placeholder origin, prints the hostname, idempotent by the distribution `Comment` (`<clusterName>-platform`).
- **`scripts/cloudfront-attach-origin.sh`** — post-install. Binds a VPC origin to the ALB the load balancer controller created and swaps the origin. No-op when already correct. **Also self-heals the documented stale-origin `curl 000` failure** via create-new / swap / delete-old, which `.kiro/steering/troubleshooting.md` describes as "NOT yet implemented".

### No platform change was needed

`insecure: true` already produced `scheme: internal` and the predictable name `<clusterName>-platform` (`gitops/addons/registry/core.yaml`). That was written for *adopting* a pre-created ALB, and it behaves identically when the controller creates it. I had expected to need a change here and did not.

## What was removed

Now at **zero references** repo-wide: `private/async-domain`, `hub:wait-for-domain` (43 lines), the `WAIT_FOR_CF` dual mode, the PID-file rejoin loop, `cf-domain.txt`, a **duplicate** `create-distribution` call site, and a ~215-line inline provisioning function.

- `workshop/create-config.sh`: **696 → 344 lines**, 0 `elbv2`/`cloudfront` calls (was 14)
- `kind-kro-ack`: 50 → 49 tasks

Because the domain is static from the start, an entire class of workaround disappears with it: "global `vars:` are evaluated once at parse time so this value is stale". `KEYCLOAK_DNS` went from a three-source fallback chain to one line.

Also unwound: `EXPOSURE_MODE` keyed off an **empty** domain to mean CloudFront, which inverts now that the domain is always set. It keys off `insecure` instead.

## Verified end to end against live AWS

A full install was run on this branch (`us-west-2`), GitHub-only, no GitLab.

| Assertion | Observed |
|---|---|
| Hostname reserved before any infrastructure | distribution created with `placeholder.invalid`; no ALB, no VPC origin |
| `task install` exit code | **0**, ~30 min |
| **ALB present the instant install returned** | **yes, zero wait** — `internal`, `active` |
| Reserved domain reached the addons | `ingress_domain_name` in `<hub>/config` matched exactly |
| `task cloudfront:attach` exit code | **0**, ~9 min (VPC origin deploy) |
| Origin swapped | distribution origin = the internal ALB, via the new VPC origin |
| attach re-run | no-op, exit 0 |
| reserve re-run | same hostname, distribution count unchanged |
| HTTP through CloudFront | `/keycloak/realms/master` **200**, `/argo-workflows` **200**, `/kargo` 301, `/jupyterhub` 302 |

The ALB-timing check was the main risk: if the ALB were not ready when install returned, the documented recipe would race. It was ready immediately.

Also measured earlier in isolation: `update-distribution` accepts replacing a `CustomOriginConfig` with a `VpcOriginConfig` on a live distribution, with the VPC origin reaching `Deployed` in 540s (554s in the full run).

### Not verified

- **`kind-crossplane` was not run.** Its blocker was structural and the structure is gone (0 domain-resolution logic remains, and no VPC is needed), but that is an argument, not a result.
- `workshop:install` end to end. It needs the CDK-provisioned GitLab EC2 instance, so `set-overlay-repo` and the GitLab-dependent tail were not exercised.
- `/backstage` returns 503 on the test cluster. It was also degraded before this branch, and an in-cluster curl straight to the ALB returns the same, so it is an application issue rather than a CloudFront one.

## Also in this PR

- **Workshop Studio decoupling**: repo root now comes from `{{.ROOT_DIR}}` rather than `WORKSPACE_PATH`/`WORKING_REPO`, so the providers work on any machine.
- **A real pre-existing bug fixed**: `( ... ) &` inside a `cmds:` block is killed when the task body exits, so `install:phase1-observability-bg` never ran its 30+ minute AMP seed on either provider. Fixed with a double-fork; verified the backgrounded job now survives.
- **Fail-fast guard**: `kind-crossplane` rejects `hub.network.vpcId` instead of silently creating a second VPC and leaving CloudFront pointed at an unusable ALB (#833).
- Null-safety for `yq` returning the string `"null"` on a missing key, which defeated `[ -z ]` guards and could write `ingress_domain_name: "null"` into the cluster secret.

## Docs

- `docs/platform/cloudfront-exposure.md` — the consumer guide: the three steps, required config, timeline, what the consumer owns versus what is shipped, and the plaintext-hop shortcut that is deliberately **not** taken.
- `cluster-providers/README.md` — records the domain obligation and makes the capability-parity rule explicit, the omission that produced #833.

## Reviewer notes

`workshop:install` calls `cloudfront:attach` **between** the platform install and `post-install`. That placement is load-bearing, not cosmetic: the ALB does not exist until the platform install creates it, and `post-install` → `idc:configure` reaches Keycloak *through* the CloudFront domain, so IDC federation cannot work while the origin is still a placeholder. Running it there also means the ~9 minute origin deploy overlaps spokes and ray instead of adding to the critical path.

An alternative branch, `feat/cluster-provider-parity-cloudfront` (11 commits, +1035/−417), solves the same problem by keeping the async design: a `domainResolver` extension point, an install phase split, and a nested go-task process to work around parse-time variable evaluation. It is fully implemented and stub-tested. This PR supersedes it — roughly two thirds of its provider changes turned out to be machinery for the ordering problem rather than the domain problem. Happy to walk through the comparison if useful.

## Follow-up found while testing, not addressed here

`task <provider>:destroy` reports exit 0 while orphaning billable resources. Deleting the cluster removes the controllers that own them, so nothing cleans up: the ALB (load balancer controller), devlake's Aurora cluster (Crossplane), and then the orphaned ALB and RDS block VPC deletion, leaving the NAT gateway and EIPs behind. Observed on this account: an Aurora `db.r5.xlarge` left running for 8 days. This is largely pre-existing, though the ALB orphan is worse here since `create-config.sh` used to create that ALB. Worth a separate issue, since repeated provision/destroy cycles will accumulate VPCs until hitting the account limit — the `VpcLimitExceeded` failure already documented in the troubleshooting guide.

Related: #833 (provider capability parity), #836 (shell-in-YAML tech debt).


hmuthusamy:
@shapirov103  looks like we have merge conflicts. changes looks good though

elamaran11 (dismissed):
Please fix the merge conflicts PR looks ok

allamand:
Fix this and i'll test it in studio

shapirov103:
@hmuthusamy @elamaran11 @allamand merge conflicts resolved. Functional validation pending, Sebastien started a test event in WS, provisioning finished.

allamand:
Results: ✅ 15 PASS · ❌ 0 FAIL · 🔎 5 observations
CloudFront config — CONFORMING (no regression):

IDE CloudFront — Deployed + Enabled, origin = IDE EC2 (Caddy /→GitLab:8929, /ide/→code-server:8111).
Platform CloudFront — Deployed + Enabled, origin = VPC Origin  → internal ALB peeks-hub-platform (active, AWS-LBC-adopted). VPC origin Deployed.
ALB target groups healthy: backstage ✅, keycloak ✅✅, argo-server ✅, kargo ✅.

allamand (approved):
Tested, LGTM

allamand:
## Setup flow — CloudFront reservation → VPC origin/ALB attach → GitOps/ArgoCD/Secrets Manager
  
  Verified against the `ssm-setup-ide-logs` (SetupIDE SSM document) output for the hub
  (`<PREFIX>-hub`, account `<ACCOUNT_ID>`, region `<REGION>`). The design deliberately **decouples**
  CloudFront distribution creation from origin attachment: the hostname is created up front,
  `task install` runs **in parallel** with CloudFront, and the origin is attached to the ALB only
  at the end.
  
  ### Timeline
  
  | Time (UTC) | Phase | What happens |
  |---|---|---|
  | `T0+00:00` | **CloudFront hostname "reserved"** | `cloudfront-reserve-domain.sh` runs `create-distribution` with an unresolvable `placeholder.invalid` origin. CloudFront returns
  `DomainName` immediately → `<PLATFORM_CF_DOMAIN>` (the **final** domain, no infra needed). Written to `config.local.yaml`; `task install` starts in parallel. |
  | `T0+25m` | **Seed Secrets Manager** | `ingress_domain_name=<PLATFORM_CF_DOMAIN>` written into `<PREFIX>-hub/config`; `<PREFIX>-hub/keycloak` and `<PREFIX>-hub/secrets` seeded (all
  `AWSCURRENT`). |
  | `T0+27m` | **ArgoCD bootstrap** | ESO + `ClusterSecretStore aws-secrets-manager` ready; `secret/<PREFIX>-hub` (ArgoCD cluster secret) created; `applicationset/bootstrap` applied → hub
  self-managing. |
  | `T0+28m–45m` | **Hub addons install (~17 min)** | LB Controller provisions the internal ALB `<PREFIX>-hub-platform` during this window. |
  | `T0+45m–54m` | **Attach CloudFront → VPC origin → ALB** | ALB `<PREFIX>-hub-platform` (**internal**, active) → distribution `<DISTRIBUTION_ID>` → **new** VPC origin `<VPC_ORIGIN_ID>`
  created for the current ALB → `Deployed` after ~8m20s → attached: `<PLATFORM_CF_DOMAIN>` → `<INTERNAL_ALB_DNS>`. |
  | `T0+54m` | **GitOps publish** | ArgoCD repo-creds `fleet-config-repo` + `gitlab-user-repo-creds` registered; `overlay_repo_url` wired into `<PREFIX>-hub/config` (ESO reconciles → activates
  the `$overlay` layer); spokes `<PREFIX>-spoke-dev` / `<PREFIX>-spoke-prod` committed to `fleet-config` (GitLab). Overlay resolves fleet-wide. |
  | `T0+56m–62m` | **Auth: IDC ↔ Keycloak (SAML/SCIM)** | Depends on CloudFront being attached: `/keycloak` reachable via CloudFront (HTTP 200) → SAML descriptor published to IDC → SCIM token
  extracted → SAML client created → user added to IDC admin group → ArgoCD access enabled. |
  
  ### Key design notes
  - **"Reserve" = real `create-distribution`, not a name reservation.** CloudFront returns the final
    `*.cloudfront.net` domain instantly; the `placeholder.invalid` origin (RFC 2606, never resolvable)
    makes it obvious no origin is attached yet. This breaks the circular dependency where the platform
    needs its ingress hostname (Keycloak realm URLs, OIDC issuer, ArgoCD/Backstage base URLs) *before*
    the ALB it points to exists.
  - **Distribution identity = `Comment` (`<PREFIX>-hub-platform`).** `cloudfront-attach-origin.sh` finds
    the same distribution by Comment; both the reserve and attach steps are **idempotent** (re-run reuses
    the existing distribution / re-points the current ALB).
  - **ALB is `internal`** (required for a VPC-origin backend), so the LB Controller adopts it in place
    instead of delete/recreating it — no stale-VPC-origin churn.
  
  ### Health assessment
  Run is clean: ALB internal/active, a **fresh** VPC origin created for the current ALB and `Deployed`
  within budget, clean attach, overlay resolving fleet-wide. No stale-VPC-origin symptoms (no curl `000`,
  no ALB recreate).

PR #898, opened 2026-09-13
release: v0.3.0-rc5 → main

allamand:
## Release: v0.3.0-rc5 → main

Promotes the **release/v0.3.0-rc5** line to `main`. Validated end-to-end on a fresh kind-kro-ack deployment (deploy + validation green; teardown one-shot with the fixes below).

## Included changes
- **fix(teardown): release Crossplane-managed DevLake RDS before hub teardown** (#895/#896) — `task destroy` now deletes the RDS claim and waits for Crossplane to reap the composed AWS resources (incl. `rds-mysql-sg-*`) before destroying the hub, so the SG is no longer orphaned in the IDE VPC.
- **fix(platform): harden system-peeks NodePool against AZ-collapse disruption** (#892/#894) — `minDomains: 2` on the critical add-ons' zone TSCs + scoped disruption budgets.
- **fix(keycloak): force replicas across ≥2 AZs** (`DoNotSchedule` + `minDomains: 2`, #892).
- **fix(jupyterhub): ALB healthcheck path** → `/jupyterhub/hub/health`.
- **fix(fleet): propagate repo URL (not just revision) to kro spoke clusters.**
- **fix(hooks): repoint helm-lint hook to relocated platform-charts.**
- **docs: align platform docs & steering with the cluster-providers architecture.**

## Validation
- Full deploy → validate → teardown cycle on kind-kro-ack (us-west-2): deployment `CREATE_COMPLETE` (hub + 2 spokes), `validate-workshop.sh` all checks passed (ArgoCD apps Synced/Healthy, 9 KRO RGDs Active, Backstage 200, GitLab 302, IDE SSM online).
- Teardown hardened one-shot: Crossplane RDS released before hub teardown (#896), with the workshop content repo's `delete-stack.sh` reaper as a defensive backstop for SG-blocked VPCs.

PR #904, opened 2026-09-15
docs(cluster-providers): restore Domain handling & Capability parity sections

allamand:
Addresses @shapirov103's review on #898 (item 1).

`035421cbe` ('docs: align platform docs & steering…') restructured the cluster-providers README Configuration section and **incidentally dropped** content that `130a166c` (#888) had added two days earlier. This restores it, fitted to the new layout (fields folded into the current `Field | Description` table; the two sections placed just before `### Configuration`). Docs-only, no functional change.

**Restored:**
- **Domain handling (required behaviour)** — a provider must *fail rather than install with an empty domain*; `domain` is a static config value, never resolved/discovered/waited-for by a provider; consumers without a domain reserve a CloudFront hostname outside the provider (link to `docs/platform/cloudfront-exposure.md`).
- **Capability parity (binding)** — providers are interchangeable behind one `config.yaml`, so a capability added to one must be added to all; where not yet true the lacking provider must **fail fast** *and* a GitHub issue must track the gap. Worked example: `hub.network.vpcId` + #833.
- **Config fields**: `hub.network.vpcId`/`subnetIds`, the 'domain must be known before install' wording, and `insecure`.

The parity contract is tracked going forward by a dedicated issue (linked below).


allamand:
Parity contract tracked going forward in #905 (ensure capability parity between cluster-providers — includes verifying whether `kind-crossplane` needs the #903 teardown backstop).

PR #906, opened 2026-09-15
fix(teardown sweep): reap orphaned <prefix>-ray-vllm-custom ECR repo

allamand:
## Problem
The Ray/vLLM module builds a custom inference image and pushes it to an ECR repo created **imperatively** by the Ray build Task (`cluster-providers/common/Taskfile.ray.yaml` → `aws ecr create-repository <prefix>-ray-vllm-custom`).

Unlike the app-layer repos (`<prefix>/rust`, `<prefix>/java`) that are **ACK-managed** via the `CICDPipeline` RGD and therefore cascade-deprovisioned on teardown (see #901), this repo has **no owning CR**. So `task destroy` never removes it, and it survives as an orphan — confirmed leftover on reused test accounts (had to be swept manually).

## Fix
Add a small ECR section to `workshop/scripts/sweep-workshop-resources.py` (run by the teardown CodeBuild after `task destroy`) that force-deletes `<prefix>-ray-vllm-custom` (images included). Non-fatal like every other sweep section; no-op when the repo is absent.

## Test
- `python3 -m py_compile` passes.
- Manually verified the equivalent `ecr delete-repository --force` clears the orphan on a reused account (repo + its 1 image removed, 0 residue after).

Scope intentionally narrow (the one known imperatively-created repo); app-layer repos remain ACK-owned and are untouched.


shapirov103 (approved):
LGTM

elamaran11 (approved):
LGTM

PR #910, opened 2026-09-16
fix(argocd-recovery): wait for operation to clear before syncing; settle before final report

### LINKED ISSUE #909: argocd recovery: terminate→sync race prints misleading "another operation is already in progress" FATA
## Summary

`scripts/recover-argocd-apps.sh` (`argocd-sync`) can print a scary `FATA[0000] rpc error: code = FailedPrecondition desc = another operation is already in progress` as its **last action**, making a run look failed even though the affected apps recover moments later.

## Root cause

In the recovery loop, a stuck app is handled with:

```
terminate_argocd_operation "$app"   # argocd app terminate-op  (asynchronous)
sleep 2
sync_argocd_app "$app"              # argocd app sync
```

`argocd app terminate-op` only *requests* termination — the application controller needs a moment to drop `.operation` and settle `.status.operationState`. After a fixed `sleep 2` the operation is often still in flight, so `argocd app sync` fails with **"another operation is already in progress"**. The kubectl fallback (`{"operation":{"sync":{}}}`) then no-ops, because it also cannot post a new operation while one is running.

The app still recovers — but via a **later ArgoCD auto-sync**, not the script's explicit sync. So a successful recovery ends on a misleading `FATA` line, and the final report is a mid-flight snapshot that can mislabel a soon-to-be-healthy app as `KO`.

## Observed

On a fresh v0.3.0 hub, `kubevela-peeks-hub` was stuck (PreSync admission hook), which cascaded to `devlake-peeks-hub` (needs `core.oam.dev` CRDs) and `grafana-dashboards-peeks-hub` (its `grafana-mysql-creds` ExternalSecret depends on the DevLake MySQL DB provisioned via KubeVela). `argocd-sync` printed the `FATA` above and reported all three `KO`; ~3 min later a KubeVela auto-sync completed, vela-core deployed, CRDs installed, and all three cascade-healed to Healthy/Synced (94/94).

## Proposed fix

1. `terminate_argocd_operation`: wait (bounded) for the operation to actually clear before returning.
2. `sync_argocd_app`: treat "another operation is already in progress" as benign — wait for the op to clear and retry once, instead of a warning + no-op patch.
3. `recover-argocd-apps.sh`: poll apps to a settled state (bounded, tunable) before the final report so it reflects the settled outcome.

PR incoming.

allamand:
Closes #909

## Problem

`scripts/recover-argocd-apps.sh` (`argocd-sync`) terminated a stuck operation then synced after only a fixed `sleep 2`. `argocd app terminate-op` is asynchronous, so the sync raced the still-in-progress operation and failed with `FATA ... another operation is already in progress`; the kubectl fallback then no-op'd (cannot post a new operation while one is running). The app still recovered — but only via a later ArgoCD auto-sync — so a **successful** run printed a misleading `FATA` as its last line, and the final report was a mid-flight snapshot that mislabelled soon-to-be-healthy apps as `KO`.

## Changes

- **`terminate_argocd_operation`** now waits (bounded, 60s) for the operation to actually clear — `.operation` absent and `operationState.phase` not `Running`/`Terminating` — before returning, so the subsequent sync no longer collides. New helper `wait_for_operation_cleared`.
- **`sync_argocd_app`** treats `another operation is already in progress` as benign: it waits for the op to clear and retries the sync once, instead of a warning + no-op kubectl patch. Other sync failures keep the existing kubectl fallback.
- **`recover-argocd-apps.sh`** polls applications to a settled state before the final report (new helper `final_verify_settle`, tunable via `FINAL_VERIFY_TIMEOUT`, default 120s), so the run reflects the settled outcome.

## Behaviour

- No more misleading `FATA` on an otherwise-successful recovery.
- Recovery drives the sync itself (deterministic) rather than relying on a later auto-sync.
- Final report reflects the settled state; fully backward compatible (additive, bounded waits, best-effort reporting).

## Testing

- `bash -n` passes on both edited scripts.
- Changes are additive/defensive; bounded timeouts guarantee termination.

_Context: surfaced on a fresh v0.3.0 hub where a stuck `kubevela` PreSync admission hook cascaded to `devlake` and `grafana-dashboards`; the run printed the `FATA` and reported `KO`, yet all three cascade-healed to Healthy/Synced (94/94) ~3 min later via auto-sync._


elamaran11 (approved):
LGTM.

PR #911, opened 2026-09-17
fix(kro-clusters): propagate resourcePrefix to spoke EksclusterWithVpc

allamand:
## Problem

On a deployment using a **custom resource prefix** (e.g. `peeks-031`), the parent `cluster-addons` ArgoCD Application stays **Degraded** because the `ack_pod_identities` ApplicationSet reports `ApplicationValidationError`:

```
application destination spec is invalid: there are no clusters with this name: peeks-hub (and 1 more)
```

## Root cause

Spoke `EksclusterWithVpc` instances **never received `resourcePrefix`**, so they fell back to the RGD schema default `"peeks"`. The nested `EksCluster` (which the `EksclusterWithVpc` RGD already forwards `resourcePrefix` to) then stamped the spoke ArgoCD cluster secret with:

- `resource_prefix: peeks`
- `hub_cluster_name: peeks-hub`   (`${resourcePrefix}-hub`)

Addons flagged `destinationHub: true` (e.g. `ack_pod_identities`) resolve their destination to `{{ default .name (index .metadata.annotations "hub_cluster_name") }}` = `peeks-hub`. Under a custom prefix the hub is registered as `peeks-031-hub`, so `peeks-hub` is not a known cluster → destination invalid → ApplicationSet error → `cluster-addons` Degraded.

Verified live on `peeks-031-hub`:

| cluster secret | resource_prefix | hub_cluster_name |
|---|---|---|
| peeks-031-hub | peeks-031 ✅ | null → falls back to `.name` = peeks-031-hub ✅ |
| peeks-031-spoke-dev | **peeks** ❌ | **peeks-hub** ❌ |
| peeks-031-spoke-prod | **peeks** ❌ | **peeks-hub** ❌ |

The RGDs are **not** at fault — `rg-eks-vpc.yaml` already threads `resourcePrefix: ${schema.spec.resourcePrefix}` into the nested `EksCluster`. The gap is purely in the **caller layer** that never sets `EksclusterWithVpc.spec.resourcePrefix`.

## Fix

- `gitops/bootstrap/clusters-kro.yaml` — thread `global.resourcePrefix` from the hub cluster secret's `resource_prefix` annotation.
- `gitops/abstractions/kro/kro-clusters/templates/clusters.yaml` — set `spec.resourcePrefix` per spoke as `per-cluster override | global | default "peeks"`.

Mirrors the spoke repoURL global-inheritance pattern from #894.

## Scope / impact

- **Default-prefix (`peeks`) deployments are unaffected** — `${resourcePrefix}-hub` coincidentally equals the real `peeks-hub`, so no workshop-participant impact.
- Only custom-prefix deployments were broken; this closes the multi-prefix isolation gap on the gitops side.
- Functionally non-blocking today (the `ack-pod-identity` addon itself is Healthy on the hub); this clears the cosmetic parent `cluster-addons` Degraded.

## Validation

`helm template` of `abstractions/kro/kro-clusters`:
- per-cluster override → renders the override value
- `global.resourcePrefix=peeks-031`, no per-cluster value → renders `peeks-031`
- global present without `resourcePrefix` → renders default `"peeks"`

## Test plan

- [ ] Deploy with a custom `--resource-prefix` (kind-kro-ack); confirm spoke cluster secrets carry `resource_prefix=<prefix>` and `hub_cluster_name=<prefix>-hub`.
- [ ] `ack_pod_identities` ApplicationSet has no `ApplicationValidationError`; `cluster-addons` reaches Healthy.
- [ ] Default `peeks` deployment unchanged.


allamand:
Added a second, closely-related resourcePrefix propagation fix in commit `1351b94b`:

**`platform-manifests` addon** — its registry `global:` block (`gitops/addons/registry/platform.yaml`) passed only `accountId`, so `templates/ray-system-iamroleselectors.yaml` fell back to `{{ .Values.global.resourcePrefix | default "peeks" }}` → the `ray-system-iam`/`ray-system-eks` IAMRoleSelectors pointed ACK at `role/peeks-cluster-mgmt-iam`, which does not exist under a custom prefix. Result: every ACK IAM Role/Policy in `ray-system` stayed `ResourceSynced=False`, the `HuggingFaceModel` RGD blocked forever on its `iamRole` node, and no model download ever started (0 pods/jobs in ray-system for 15h on peeks-031).

Fix threads `global.resourcePrefix` from the `resource_prefix` cluster-secret annotation (same pattern already used by the `multi-acct` and `platform-manifests-post` entries) and drops a duplicate `clusterName` line.

Validated live on peeks-031 (custom prefix `peeks-031`).

allamand:
Added a third resourcePrefix-propagation fix in commit `19e52b50`, one layer deeper (the app/CI/CD RGDs themselves):

**`AppmodService` / `CicdPipeline` / `S3Bucket` RGDs** embedded IAMRoleSelectors with a **literal** `role/peeks-cluster-mgmt-<svc>` ARN (`gitops/addons/charts/kro/resource-groups/manifests/{appmod-service,cicd-pipeline/cicd-pipeline,rg-s3-bucket}.yaml`). Under a custom prefix that role doesn't exist, so ACK can't manage the app's IAM/EKS/DynamoDB/ECR/S3 resources in the team namespace — the same `ResourceSynced=False` → RGD-hangs-on-iamRole failure as the ray-system and spoke cases.

- `appmod-service.yaml` / `rg-s3-bucket.yaml`: added `resourcePrefix` (`default="peeks"`) to the schema and switched the ARN to `${schema.spec.resourcePrefix}`.
- `cicd-pipeline.yaml`: the schema **already** had `spec.aws.resourcePrefix` (`default="peeks"`) but the ARN ignored it — now uses `${schema.spec.aws.resourcePrefix}`.

Backward-compatible (`default="peeks"` keeps default-prefix workshops identical).

⚠️ **Follow-up (separate layer, not in this PR):** to *fully* unblock a custom prefix for the app-deploy modules, the **callers** that instantiate these RGDs must pass `resourcePrefix`. Today `applications/{rust,java,next-js}/…/application.yaml` and the backstage s3-bucket skeleton pass `accountId` but not `resourcePrefix`, and the workshop content only `export AWS_ACCOUNT_ID` before `envsubst`. Wiring that requires editing participant-facing app manifests + content (a new `RESOURCE_PREFIX` env/placeholder), so it's better tracked as a content-layer change than folded into this platform-gitops PR. This commit is the necessary platform-side foundation for it.

allamand:
Added the caller layer for the app-deploy path in commit `26cb3f0c`:

**`AppmodService` instances now pass `resourcePrefix: "${RESOURCE_PREFIX}"`** (rust `deployment/templates/kro/application.yaml`, next-js `deployment/dev/application.yaml`). It's resolved by the **same envsubst step** that already substitutes `${AWS_ACCOUNT_ID}` when the participant runs `envsubst < deployment/.../application.yaml | kubectl apply`.

Verified that `RESOURCE_PREFIX` is exported in the IDE shell env by the CDK bootstrap — `~/.bashrc.d/platform.sh` (interactive) and `/etc/profile.d/workshop.sh` (login), value from the CFN `ResourcePrefix` param. So it resolves to the real prefix (e.g. `peeks-031`) on a custom deployment and to `peeks` on a default one. Combined with the RGD schema `default="peeks"`, default-prefix workshops are byte-identical.

Note: `envsubst` does not honor `${VAR:-default}`, so the "peeks default" comes from the RGD schema default + RESOURCE_PREFIX always being exported (not from manifest syntax). A belt-and-suspenders `export RESOURCE_PREFIX=${RESOURCE_PREFIX:-peeks}` next to the envsubst call in the workshop content (platform-engineering-on-eks repo) would guard the empty-env edge.

**Deliberately out of scope here** (different mechanism, separate follow-up): the **java** kro manifest is static (hardcodes `clusterName: peeks-spoke-dev`, no envsubst) and the **backstage scaffolder skeletons** (S3Bucket / CICDPipeline) use `${{values.*}}`; both would need their own parameterization (and java hardcodes other peeks-prefixed values), so a blind `${RESOURCE_PREFIX}` add there would break them.

applications/java/deployment/templates/kro/application.yaml
```diff
@@ -11,6 +11,7 @@ spec:
   targetPort: 8080
   serviceAccount: java-service-account
   clusterName: peeks-spoke-dev
```
shapirov103, line 13: i see we hardcode clustername and prefix  - it is fixed for the workshop. maybe we should address it should just be envName placeholder for cluster name.  Also our resource prefix while better than nothing does not segregate resources provisioned in the same account  - account and region level (e.g. roles, dynamodb, etc.).

I think the PR is an improvement but curious to hear your thoughts on the env segregation - I would say envname should be the prefix for resource for clean separation.

shapirov103 (commented):
Please see my comment.

applications/java/deployment/templates/kro/application.yaml
allamand: This is a place holder already seded by the instructions. It works also without upgrade when using defaults that why I didn't put it place holder spécifique.

Normally with prefix we should be able to deploy several instances in same account, at least that is my goal

allamand: IAM roles.. should inherit also of the resource-prefix, unsure for dynamodb, needs to check

allamand: Validated this concern live on a fresh custom-prefix deploy (KRO/ACK path, prefix `peeks-032`). Confirmed: the prefix propagates to the pipeline / cluster-mgmt layer but **not** to the app data-plane resources.

| Resource | Layer | Observed name | Prefixed? |
|---|---|---|---|
| CI/CD IAM role | pipeline | `peeks-032-rust-role` | ✅ |
| CI/CD IAM policy | pipeline | `peeks-032-rust-ecr-policy` | ✅ |
| ECR repos | pipeline | `peeks-032/rust`, `peeks-032/rust/cache` | ✅ |
| DynamoDB table | app | `rust-app-table-dev` | ❌ |
| App pod-identity role | app | `rust-microservice-dev-role` | ❌ |

The only prefix-bearing marker on the table was the ACK-injected `eks:eks-capability-arn` tag (identifies the creating cluster capability, not app-level segregation).

**Root cause:** the `AppmodService` RGD's `resourcePrefix` only drives the IAMRoleSelector cluster-mgmt ARNs. The table name comes from the caller's `components.dynamodb.tableName` (hardcoded `rust-app-table-<environment>`) and the app role is `${schema.metadata.name}-role` — neither used the prefix. So two envs in the same account+region both create `rust-app-table-dev` / `rust-microservice-dev-role`, and `adopt-or-create` silently cross-adopts the first env's table.

**Follow-up commit** (f23863f3) addressing it on the KRO/ACK path:
- `AppmodService` RGD: app pod-identity role → `${schema.spec.resourcePrefix}-${schema.metadata.name}-role`
- rust KRO caller: DynamoDB `tableName` + the app's `TABLE_NAME` env → `${RESOURCE_PREFIX}-rust-app-table-<environment>` (kept in sync so the app still reads the table it provisions)

With the default prefix this yields `peeks-rust-app-table-dev` / `peeks-rust-microservice-dev-role`, consistent with the rest of the platform (`peeks-hub`, `peeks-cluster-mgmt-*`).

**Out of scope in this commit:** the KubeVela/Crossplane alternative path (`applications/rust/deployment/templates/kubevela`) derives table/role names from component/appName via CUE with no `resourcePrefix` parameter — a larger, separate change, and not the default deploy path.

allamand: Follow-up commit ca42df0a extends the same prefixing to the **KubeVela/Crossplane** deploy path so both paths stay in parity:

- `dp-service-account` ComponentDefinition: new `resourcePrefix` param (default `peeks`); app IAM role → `${resourcePrefix}-${name}-iam-role` (Role, pod-identity `roleArnRef`, and `RolePolicyAttachment` `roleRef`/`policyArnRef` all updated consistently).
- `component-iam-policy` trait: new `resourcePrefix` param; policy → `${resourcePrefix}-${appName}-${component}-iam-policy` (both branches).
- rust KubeVela caller: DynamoDB `tableName` + `TABLE_NAME` env prefixed, and `resourcePrefix` passed to the service-account component and the policy trait.

Both deploy paths (KRO/ACK and KubeVela) now segregate the app table + IAM role/policy by prefix. Not yet re-validated on a fresh deploy (immutable/adopt-or-create means an in-place retrofit won't show it cleanly).

allamand: ✅ **Dual-path validation on a fresh env — both paths now inherit the prefix on the app data-plane.**

Redeployed a **fresh** environment (prefix `peeks-033`, us-west-2) whose IDE clones this branch at HEAD, so the deployed definitions carry both fixes: the `AppmodService` RGD (KRO/ACK, `${resourcePrefix}-${name}-role`) and the KubeVela CUE definitions `dp-service-account` / `component-iam-policy` (`resourcePrefix` parameter threaded through). A fresh env is required — an in-place resync won't show the new names because of ACK field immutability / adopt-or-create.

**Baseline on the shared account (living proof of the collision this fixes):** `0` DynamoDB tables, but leftover **unprefixed** rust roles from a co-located default-prefix env — `rust-microservice-dev-role` (created via the KRO/ACK path) and `rust-service-account-iam-role` (created via the KubeVela/Crossplane path). Exactly the account/region collision the concern predicted.

### Path 1 — KRO/ACK (deployed on spoke-dev)
| Resource | Actual name | Prefixed |
|---|---|---|
| CI/CD IAM role | `peeks-033-rust-role` | ✅ |
| CI/CD IAM policy | `peeks-033-rust-ecr-policy` | ✅ |
| ECR repos | `peeks-033/rust`, `peeks-033/rust/cache` | ✅ |
| **DynamoDB table** | `peeks-033-rust-app-table-dev` | ✅ |
| **App pod-identity role** | `peeks-033-rust-microservice-dev-role` | ✅ |

App is **functional**: Rollout 2/2 available, `GET /collection/FRONT_PAGE` → HTTP 200 (full product catalog), independent DynamoDB scan = 26 items (real read **and** write via `TABLE_NAME=peeks-033-rust-app-table-dev`).

### Path 2 — KubeVela/Crossplane (deployed on the hub, `environment=prod` to avoid colliding with the KRO/ACK dev table)
| Resource | Actual name | Prefixed |
|---|---|---|
| **DynamoDB table** | `peeks-033-rust-app-table-prod` | ✅ |
| **App IAM role** | `peeks-033-rust-service-account-prod-iam-role` | ✅ |
| **App IAM policy** | `peeks-033-rust-application-rust-dynamodb-prod-iam-policy` | ✅ |

App is **functional**: the 3 Crossplane Managed Resources Synced/Ready, Rollout 2/2, `GET /collection/FRONT_PAGE` → HTTP 200, prod table scan = 26 items.

### Verdict
The concern is resolved on **both** deployment paths: the DynamoDB table **and** the app IAM role/policy now inherit the `peeks-033-` prefix (before the fix they were bare `rust-app-table-dev` / `rust-microservice-dev-role`). Two concurrent envs in the same account/region no longer collide, and the app runs end-to-end in both cases.

applications/java/progressive-app.yaml
```diff
@@ -11,6 +11,7 @@ spec:
   targetPort: 8080
   serviceAccount: progressive-service-account
   clusterName: peeks-spoke-dev
+  resourcePrefix: peeks
```
hmuthusamy, line 14: not adding prefix in this file?

shapirov103 (approved):
LGTM

elamaran11 (approved):
LGTM

applications/java/progressive-app.yaml
allamand: checking, thanks

allamand: Thanks Hari, I'm opening a complementary PR

PR #913, opened 2026-09-18
fix(kro-ack): honour configured fleetRepo, matching kind-crossplane

shapirov103:
## Problem

`kind-crossplane` reads `fleetRepo.url` / `.revision` / `.basepath` from config and falls back to `repo.*`. `kind-kro-ack` defined no `FLEET_REPO_*` vars at all and hardcoded its fleet sites to `REPO_*`, so it **silently ignored** a configured `fleetRepo`.

A consumer repo that owns its own fleet content (`fleet/members`, `fleet/spoke-values`, overlays) while consuming platform addons from here therefore got a hub cluster secret carrying:

```
fleetRepoURL: https://github.com/aws-samples/appmod-blueprints.git
```

even though its generated config set `fleetRepo.url` to its own repo. Because `gitops/bootstrap/clusters-crossplane.yaml:34` reads `repoURL: '{{ .metadata.annotations.fleetRepoURL }}'` with no overlay fallback, the generator globbed the **platform** repo's `fleet/spoke-values/tenants/*/crossplane-clusters/values.yaml`. A spoke declared only in the consumer repo was never seen, and **no error was raised** — the consumer simply could not declare spokes on this provider.

This is the capability-parity case the provider contract calls out: a field that means different things depending on `clusterProvider`, failing silently rather than failing fast.

## Changes

**1. `fix(kind-kro-ack)` — honour the configured fleetRepo**

Adds the three vars, expressions copied verbatim from `kind-crossplane`, and uses them at the two fleet sites. `addonsRepo*`, `platformRepo*` and `workloadRepo*` stay on `REPO_*`:

- the hub claim's `gitops` block (inherited by spokes through the RGD, so it matters twice)
- the cluster-secret / Secrets Manager metadata JSON

Also fixes a latent bug found while doing this: the seed cluster secret **already** substituted `{{.FLEET_REPO_URL}}` et al, but those vars did not exist. go-task renders an undefined var as the empty string, so that secret was being created with `fleetRepoURL: ""`. It now carries the real value.

`config.schema.json` documents `fleetRepo` as optional. Both providers read it; neither declared it, so a consumer had no way to discover it. All subfields are optional to match the per-field fallback.

**2. `fix(kro)` — publish subnet ids and cluster SG from the spoke RGD**

`rg-eks.yaml` hardcoded `cluster_security_group_id: ""` and published no `aws_private_subnet_ids`, while the crossplane `platform-cluster` Composition patches `aws_vpc_id`, `aws_private_subnet_ids` **and** `aws_cluster_security_group_id`. Since `crossplane-env-config` renders its `EnvironmentConfig` from those annotations (`gitops/addons/registry/core.yaml`), a kro-provisioned spoke came up with an empty `clusterSecurityGroupId` and no `privateSubnetIds` — so composed resources behaved differently depending on which provider created the spoke. Both are now published.

## Compatibility note (reviewers: this is the one risk)

`${ekscluster.status.clusterSecurityGroupID}` requires ACK **eks-controller ≥ v1.13.0** (published 2026-04-25). The field is absent in ≤ v1.12.0 and present continuously through v1.23.0. ACK runs here as an AWS-managed EKS Capability, so its version is not pinned in this repo and I could not read it without a live cluster. Worth confirming against the target cluster:

```bash
kubectl get crd clusters.eks.services.k8s.aws -o json \
  | jq -r '.spec.versions[0].schema.openAPIV3Schema.properties.status.properties
           | has("clusterSecurityGroupID")'   # must print true
```

The two commits are deliberately separate so commit 2 can be dropped independently if this turns out to be a problem; commit 1 has no such dependency.

## Backwards compatibility

With no `fleetRepo` block the three vars resolve **byte-identically** to the previous `REPO_*` values, including the `"<basepath>/fleet/"` concatenation. Fallback is per-field: setting only `fleetRepo.url` leaves revision and basepath on `repo.*`.

One behaviour change does occur on the no-`fleetRepo` path, and it is the latent-bug fix above: the seed cluster secret's `fleetRepoURL` goes from `""` to the real URL.

The pre-existing double slash in the claim's `fleetRepoBasePath` (`gitops/` + `/fleet/` → `gitops//fleet/`) is **preserved deliberately** — normalizing it would change behaviour on the no-`fleetRepo` path. The annotation the appsets actually concatenate is correct at a single slash.

## Verified by running

- go-task renders an undefined var as empty (this is what established the seed-secret bug)
- all three config shapes: `fleetRepo` absent → falls back identically; set → consumer repo; partial → per-field fallback
- the basepath concatenations, including that the appset-consumed annotation yields a single slash
- YAML parses on all three files; the embedded JSON in both secret variants parses, inner and outer
- the ACK CRD field across 12 upstream tags with release dates, to establish the v1.13.0 floor
- merged current `main` (45 commits) and re-verified all of the above; main independently touched both files, with no semantic overlap

## Verified by reading only

- that every fleet appset selects `fleet_member: control-plane`, i.e. the hub secret this change fixes
- that `clusters-crossplane.yaml:34` lacks the overlay fallback that `clusters-kro.yaml:39` has
- that the crossplane Composition patches the three annotations the rg-eks change now mirrors

## Not verified

No live install was run from this branch. The RGD is delivered as `type: manifest` (raw), not Helm, so there is no render step to exercise; `kubectl --dry-run=client` needs the KRO CRD. Specifically unproven at runtime: that KRO resolves the new `clusterSecurityGroupID` reference, and that a spoke declared only in a consumer repo is globbed and provisions. A reviewer with a live hub can confirm the fix cheaply without waiting for a spoke:

```bash
aws secretsmanager get-secret-value --secret-id <hub>/config --region <region> \
  --query SecretString --output text | jq -r '.metadata|fromjson
  | {fleetRepoURL, fleetRepoRevision, fleetRepoBasepath, addonsRepoURL}'
```

`fleetRepo*` should reflect the consumer repo while `addonsRepoURL` stays on the platform repo.


hmuthusamy (approved):
LGTM

elamaran11 (approved):
LGTm

PR #914, opened 2026-09-19
fix: make spoke cluster names arbitrary, not resourcePrefix-spoke-*

shapirov103:
## Problem

A spoke cluster only provisioned if its name happened to match `<resourcePrefix>-spoke-*`. The convention was undocumented, unvalidated, and enforced by IAM at a point where a VPC, two public and two private subnets, an IGW and a NAT gateway had already been created.

Observed on a `kind-kro-ack` hub (`peeks-hub`, `resourcePrefix: peeks`) with a spoke declared as `oap-dev`. Networking provisioned fine, then `CreateCluster` failed permanently:

```
AccessDeniedException: User: .../peeks-cluster-mgmt-eks/aws-go-sdk-... is not authorized to
perform: iam:PassRole on resource: .../role/oap-dev-cluster-role because no identity-based
policy allows the iam:PassRole action
```

ACK was assuming `peeks-cluster-mgmt-eks` correctly via CARM, so role *selection* worked. CloudTrail showed five identical attempts over ~80s, so this was a real denial rather than a propagation race.

`project.md` already states that cluster names are customer-supplied and arbitrary and must never be derived from or enforced against `resourcePrefix`. This makes the code match that.

## Why the failure had five parts

1. **PassRole scoped by role NAME** (`role/<prefix>-*`). The resource graph names all 12 of its roles after the CLUSTER, so `oap-dev-cluster-role` was not passable. The conditioned wildcard that existed covered only `pods.eks.amazonaws.com`, which is not what the cluster role (`eks`), node role (`ec2`) or capability roles (`capabilities.eks`) are passed to.
2. **The cluster-mgmt TRUST policy scoped by role name** (`<prefix>-spoke-*-ack-capability-role`). Fixing PassRole alone only moves the failure later: once a non-conforming spoke exists, its own `<cluster>-ack-capability-role` cannot assume `<prefix>-cluster-mgmt-*`.
3. Every role the graph creates is cluster-named, not prefix-named (12 of 12).
4. **Teardown only found prefixed resources**, so a non-conforming spoke's VPC, NAT gateway, EIP, ALBs and IAM roles were left behind silently.
5. **The kro RGDs emitted no ownership tags**, so tag-based cleanup could not cover them either, while the crossplane Composition already tagged.

All six claims in the report were verified against `main` before changing anything.

## Option chosen: make names genuinely arbitrary

Decoupling resource naming from cluster naming (roles as `<prefix>-<cluster>-<purpose>`) was evaluated and **rejected on evidence**: `UpdateClusterConfig`'s request syntax has no top-level `roleArn`, only `computeConfig.nodeRoleArn`. An existing cluster's service role cannot be repointed, so renaming would force ACK to recreate live clusters. That approach is greenfield-only.

Keeping the convention and merely validating it was rejected because it permanently constrains customer naming, which is the objection, and leaves teardown blind for every name.

**Nothing is renamed here, so existing clusters are unaffected.**

## Changes

**`fix(providers)` — authorize by target service, not by name.** PassRole is conditioned on `iam:PassedToService` across the four services that are the actual trust targets (`eks`, `ec2`, `capabilities.eks`, `pods.eks`). The trust policy's third pattern becomes `*-ack-capability-role`, still constrained to this account (the Principal is the account root) and to the platform-specific suffix. Both blocks were duplicated across `kind-kro-ack` and `kind-crossplane` and are changed identically.

**`feat(kro)` — ownership tags.** `platform.gitops.io/prefix` and `platform.gitops.io/cluster` on the EKS cluster and the VPC, using the same keys the crossplane Composition already patches, so one discovery query covers either provider. The VPC tag is the one that matters: teardown deletes the cluster first, so by sweep time there is no cluster left to read a name from and the VPC anchors what remains.

**`fix(teardown)` — discover spokes by tag.** Both the spoke list and the spoke-VPC lookup now use the ownership tag, and spoke names may also be passed on the command line. Discovery is **strictly additive** — the conventional names are always included and a discovery-API failure falls back to the old list rather than sweeping nothing. Section 10 now explicitly refuses the hub VPC, which carries the same tag; without that guard, tag-based discovery would have pulled the hub into the spoke sweep.

**`fix(clusters)` — fail fast at declaration time.** With authorization no longer keyed on the name, two real limits remain, enforced by a Helm `fail` in the templates that render the instance and the claim. That is the chokepoint every declaration path funnels through, workshop tasks and a consumer writing fleet JSON alike, and a failure means the Application does not render so nothing is created.

| Limit | Why |
| --- | --- |
| DNS-1123 label | The name is used verbatim as a Kubernetes namespace in 21 places across the resource graphs. EKS would accept uppercase and `_`; the namespace will not. |
| ≤ 34 (kro) / ≤ 44 (crossplane) | Roles are `<name><suffix>` against IAM's 64-char limit. Longest suffix is `-cloudwatch-observability-role` (30) on the kro path, `-kro-capability-role` (20) on crossplane. Each chart enforces its own real bound. |

The kro check sits inside the `targetCluster` gate so one bad declaration fails only its own Application, not the whole fleet.

## A gap found while verifying

Rendering showed `spec.resourcePrefix: null` on the kro path — the chart never passed it. The new tag would have fallen back to the RGD's schema default `"peeks"`, so a differently-prefixed install would have tagged every spoke as belonging to `peeks` and the tag-based sweep would have found nothing. Fixed by plumbing the hub secret's `resource_prefix` annotation through the appset and chart, as `fleet-secrets.yaml` already does. The key is omitted when unavailable, so the schema default still applies rather than an empty-string tag.

## Verified by running

- `helm template`, 9 cases: arbitrary names accepted on both providers; uppercase and underscore rejected; the 34/35-character boundary exact in both directions; the legacy `peeks-spoke-dev` still accepted; both charts still render against their shipped values files
- sweeper discovery against stubbed clients, 8 cases: arbitrary tagged spokes found, legacy untagged unchanged, discovery-API failure falls back, argv escape hatch, hub excluded by id and by name, dedupe across both filters, and **a cluster tagged with a different prefix is not swept**
- both IAM policy documents extracted and parsed as JSON in both providers, with the resulting `PassedToService` list and `PrincipalArn` patterns printed and compared for parity
- all touched YAML parses; the sweeper compiles; zero `spoke-*` name patterns remain in either provider's IAM

## Verified by reading only

That all 12 RGD roles are cluster-named and which service each trusts; that the sweeper's `spokes` list is consumed only as cluster names; that the appsets select `fleet_member: control-plane`; that `resource_prefix` is an established cluster-secret annotation.

## Not verified

No live provisioning or teardown was run. Three acceptance criteria remain open until someone exercises them on a cluster:

- an arbitrarily-named spoke reaching a healthy EKS cluster, including its own ACK capability role assuming `<prefix>-cluster-mgmt-*`
- destroy removing everything for such a spoke
- a kro-provisioned spoke's `EnvironmentConfig` matching a crossplane-provisioned one

The IAM changes take effect on the next `hub:create-mgmt-roles` run; on an existing hub that task must be re-run for the new policy to apply.

## Out of scope

Two defects reported alongside this are left for separate work: the Crossplane v2 `Usage` namespace incompatibility in the platform-cluster Composition, and orphaned CRDs left behind when `kro_manifests_hub` is disabled and re-enabled.


allamand (commented):
Seems very good proposal. I'll test it

punkwalker (approved):

allamand:
## End-to-end validation of #914 — full deploy + arbitrary-named 3rd spoke, confirmed live via `kubectl`

Validated the full PEEKS platform on a clean account with a **custom resource prefix**, then exercised the core intent of this PR by provisioning a **3rd spoke with a truly arbitrary name (`oap-test`)** through the normal declarative path and connecting to it directly.

**Environment:** `eu-central-1`, resource prefix `peeks-914`, platform pulled from `fix/arbitrary-spoke-names`. Stack `CREATE_COMPLETE`, `deploy.sh` rc=0, `task install` ran ~78 min with **no exit 201** (the accessEntry ordering race did not bite).

### 1. Base deploy (prefix `peeks-914`)
- Spokes `peeks-914-spoke-dev` + `peeks-914-spoke-prod` **ACTIVE** (clusters + VPCs); hub `peeks-914-hub`; 7 nodes; 48 ArgoCD apps (41 Healthy); 10 KRO RGDs Active.
- Ownership tags on **all EKS clusters and both spoke VPCs**: `platform.gitops.io/prefix=peeks-914` + `platform.gitops.io/cluster=<name>` — value is `peeks-914`, not the schema default `peeks`, confirming `resourcePrefix` propagation.

### 2. Arbitrary-named 3rd spoke `oap-test` (the crux of this PR)
Created via the same declarative writes the Backstage `eks-cluster-template` makes to the fleet-config overlay repo — 4 files:
- `configs/multi-acct/values.yaml` → `oap-test: "<account-id>"` (the **CARM namespace→account mapping** — the piece that wires ACK to assume the prefix-scoped `*-cluster-mgmt-*` role instead of the base capability role)
- `kro-clusters/values.yaml` → `oap-test` spec (CIDR `10.170.0.0/16`, `resourcePrefix: peeks-914`)
- `clusters/oap-test.json` → registers with the `clusters-kro` ApplicationSet generator
- `gitops/fleet/members/oap-test/values.yaml` → registers `oap-test` as an ArgoCD cluster (`fleet-secrets` appset), env profile `dev` for the full addon set

Reconcile chain worked as designed → KRO/ACK provisioned real AWS resources under the arbitrary name:

| Check | Evidence |
|---|---|
| EKS cluster (name ≠ `<prefix>-spoke-*`) | `oap-test` **ACTIVE**, v1.35 |
| VPC | **available**, `10.170.0.0/16` |
| Ownership tags (EKS + VPC) | `platform.gitops.io/cluster=oap-test`, `platform.gitops.io/prefix=peeks-914` |
| Per-cluster IAM roles for the arbitrary name | 11 created: `oap-test-cluster-role`, `oap-test-cluster-node-role`, `oap-test-ack-capability-role`, `oap-test-argocd-role`, `oap-test-kro-capability-role`, … |
| Authorization model | `ec2:CreateVpc` etc. granted via the **shared, prefix-scoped** `peeks-914-cluster-mgmt-*` roles (trust allows `peeks-914-*` / `*-ack-capability-role`, **name-agnostic** — not `<prefix>-spoke-*`) |
| ArgoCD apps targeting `oap-test` | **25/25 Synced/Healthy** |

### 3. Direct in-cluster proof (`kubectl` against the `oap-test` API server)
```
$ kubectl get nodes -o wide
NAME       STATUS   ROLES    AGE     VERSION               INSTANCE
node-a     Ready    <none>   3h16m   v1.35.6-eks-bca9cf6   c6a.xlarge (eu-central-1a)
node-b     Ready    <none>   3h15m   v1.35.6-eks-bca9cf6   c6a.xlarge (eu-central-1b)
node-c     Ready    <none>   3h8m    v1.35.6-eks-bca9cf6   c6a.xlarge (eu-central-1b)
node-d     Ready    <none>   3h16m   v1.35.6-eks-bca9cf6   c6a.xlarge (eu-central-1b)
```
4 Bottlerocket EKS Auto Mode nodes Ready; **0 non-Running pods** across crossplane-system 16/16, kube-system 8/8, cert-manager 6/6, external-secrets 6/6, kyverno 6/6, kube-prometheus-stack 5/5, argo-rollouts 3/3, vela-system 2/2.

### Conclusion
A spoke named `oap-test` — outside the old `<prefix>-spoke-*` pattern — received all its IAM roles, access entries, pod-identities, addons, **and Auto Mode nodes that join the cluster**. The old name-scoped `iam:PassRole`/trust/access-entry authorization could never have granted these to an arbitrary name. **#914 works end-to-end.** 👍

allamand (approved):

PR #915, opened 2026-09-20
fix(teardown): make the workshop sweep name-agnostic and eliminate residue

allamand:
## Problem

The Delete-time sweep (`workshop/scripts/sweep-workshop-resources.py`, run by the `ClustersStackDeploy` CodeBuild after `task destroy`) leaves orphaned residue on reused accounts. PR #914 (arbitrary spoke names) widens the gap: several sections key on the hard-coded `<prefix>-spoke-dev` / `<prefix>-spoke-prod` names, so a spoke created with an arbitrary name (e.g. `oap-test`) — and its VPC, IAM roles, secrets and log groups — escapes the sweep entirely. There is also no S3 sweep at all, so the Ray/vLLM model-cache bucket (2+ GB) persists and keeps billing.

Observed leftovers after a full teardown: spoke VPCs, unassociated EIPs, the `*-ray-models` S3 bucket + deploy-staging bucket, arbitrarily-named-spoke IAM roles, and stale log groups.

## Changes

- **Name-agnostic spoke discovery.** New `_discover_spokes()` returns the union of the legacy name pattern and every non-hub EKS cluster carrying the #914 ownership tag `platform.gitops.io/prefix=<prefix>`. All downstream sections consume this list, so arbitrarily-named spokes are reaped too. Fully backwards-compatible — on pre-#914 environments the tag is simply absent and the legacy names are used.
- **New S3 section (8c).** Reaps imperatively-created, unowned buckets under the resource prefix — the Ray/vLLM `<prefix>-ray-models-<acct>` (and `<hub>-ray-models-<acct>`) model cache and the self-serve `<prefix>-workshop-*` deploy-staging bucket — the same unowned class as the ECR repo already handled in 8b. Empties all object versions + delete markers first. **Prefix-scoped and region-checked**, so shared bootstrap buckets (`cdk-hnb659fds-*`, `ws-assets-*`) are never matched.
- **IAM (7).** Also reap roles/policies named after any discovered spoke (`oap-test-cluster-role` etc. do not start with the prefix); policies are now paginated.
- **Secrets Manager (8).** Include discovered spoke names so an arbitrary spoke's `<cluster>/config` secret is force-deleted.
- **CloudWatch Logs (9).** Broaden to `/aws/eks/*` **and** `/aws/containerinsights/*` for the hub and all discovered spokes, plus `/aws/lambda/<prefix>-*` custom-resource log groups.
- **Elastic IPs (13).** Also match the ownership tag and tag keys (not only tag values); still only releases UNASSOCIATED addresses.
- **Orphan VPC reaper (14).** Add the `platform.gitops.io/prefix` tag to VPC discovery (catches any owned VPC regardless of name) and extend the ENI-detach retry budget from ~2 min to ~4 min (Auto Mode / KRO ENIs occasionally take longer than the old budget, which was leaving spoke VPCs behind).

## Safety

- Every deletion stays scoped to the resource prefix or a discovered owned spoke; the sweep still swallows all exceptions and exits 0 (never blocks the CFN delete).
- S3 matching is prefix-scoped + region-scoped and explicitly cannot hit `cdk-hnb659fds-*` / `ws-assets-*`.
- EIP release remains guarded to UNASSOCIATED addresses only.

## Testing

- `python3 -m py_compile workshop/scripts/sweep-workshop-resources.py` passes.
- Motivated by a real end-to-end teardown where an arbitrarily-named 3rd spoke plus the ray-models bucket / EIPs / spoke VPCs had to be cleaned by hand. Recommend a full provision → arbitrary-name spoke → destroy loop with an AWS residue check before merge.


shapirov103 (approved):
This file is growing quick. and there is no convention that if someone adds a resource to the solution that they will reflect the removal. I think we need a better strategy. As a stop gap for now I am approving.

allamand:
Yes please let's discuss if you see a better way to handle this

PR #917, opened 2026-09-22
fix(idc): make IDC↔Keycloak federation resilient to transient network errors

allamand:
## Problem

During workshop install, the IDC ↔ Keycloak federation step (`configure_identity_center.py`, run via `workshop` `post-install`) intermittently fails with Playwright `net::ERR_NETWORK_CHANGED` on the **first** `page.goto` to the AWS console sign-in URL. When it fails, IDC is left unconfigured and SCIM provisioning is skipped — surfacing downstream as `⚠ SCIM admin group not found` and broken ArgoCD SSO.

Observed in `SetupIDE` SSM logs:
```
Signing into AWS Console...
Error: Page.goto: net::ERR_NETWORK_CHANGED at https://signin.aws.amazon.com/federation?...
⚠ IDC configuration failed. Run manually: task workshop:idc:configure
⚠ SCIM admin group not found.
```
Immediately followed by `Waiting for vLLM image build...` (multi-GB download/push).

## Root cause (two compounding issues)

1. **Concurrency.** `workshop/Taskfile.yaml` ran `ray:setup` — which launches a multi-GB `docker build && docker push` of the vLLM image in the **background** (`nohup ... &`, see `cluster-providers/common/Taskfile.ray.yaml`) — **before** `post-install`. The image push saturated/reconfigured the IDE network while headless Chromium was navigating, so Chromium aborted with `ERR_NETWORK_CHANGED`.
2. **No resilience.** A single transient `page.goto` failure aborted the whole run (fail-fast), with no retry.

## Fix

- **Reorder** the workshop install so `post-install` (`idc:configure`) runs **before** `ray:setup`. The background image push can no longer race the Playwright navigation.
- **Add `goto_with_retry()`** in `configure_identity_center.py`: retries `page.goto` on transient network errors (`ERR_NETWORK_CHANGED`, `ERR_TIMED_OUT`, `ERR_CONNECTION_RESET`, ...) with exponential backoff, and re-raises non-transient errors immediately. All four `page.goto` call sites now use it.

## Testing

- Unit-tested `goto_with_retry`: retry-then-succeed on transient error, immediate propagation of non-transient errors, raise-after-exhaustion. All pass.
- `python3 -m py_compile` and YAML/`task --list` parse checks pass.
- Re-ran `task idc:configure` end-to-end on a live workshop hub: IDC now federated with Keycloak, **6 users + 4 groups** provisioned via SCIM, admin group membership synced for ArgoCD SSO.


scripts/configure_identity_center.py
shapirov103, line 1: This file is atrocious already  - getting to 1000 lines of code. How much of that code is related to retry, circuit breaker and other resilience pattterns - looks like the essence of this PR is to avoid failure on transient issues. 
If we go down this path we should start applying standard python mechanisms to handle standard transient issues, like https://github.com/jd/tenacity and such.

shapirov103 (approved):
I am approving now, however please see my comment about the IDC python file. I can accept a 30 line python script or a 100 line script. A 1000 line - it is more than a script at this point. we need to start thinking about maintenance.

scripts/configure_identity_center.py
allamand: Very good point let's crate an issue to handle this

PR #918, opened 2026-09-22
fix(kubevela): harden admission-webhook bootstrap against the vela-core chicken-egg

allamand:
## Problem

On a fresh cluster, the vela-core admission webhooks are registered **before** the vela-core Deployment is Ready. With the upstream default `failurePolicy=Fail`, the webhooks then **reject every create/update they intercept** — including the objects ArgoCD needs to finish its own sync — so the ArgoCD op wedges on the cert/admission hooks and `vela-system` stays empty. Classic chicken-egg; observed intermittently on `spoke-dev` while `hub`/`prod` won the race.

## What was already in place (verified, left as-is)

- **Retry** — the `kubevela` Application already has `syncPolicy.retry.limit=-1` with exponential backoff (5s→10m). A wedged op is already retried; no change needed.
- **Sync-wave ordering** — `cert-manager` is wave **2**, `kubevela` is wave **3** (verified live on all three clusters), so cert-manager is Ready before kubevela. No change needed.

## Fix (registry `valuesObject.vela-core.admissionWebhooks`)

- **`failurePolicy: Ignore`** — a not-yet-ready webhook can no longer block object creation, so the bootstrap completes. Trade-off: OAM validation is best-effort while vela-core is briefly unavailable (acceptable vs. a hard bootstrap deadlock). Already covered by the Application's `ignoreDifferences` (`.webhooks[].failurePolicy`) so the apiserver/vela-core cannot flap it into permanent drift.
- **`certManager.enabled: true`** — issue the webhook serving cert via **cert-manager** (a managed addon on every cluster, already Ready at wave 2) instead of the fragile PreSync `kube-webhook-certgen` Job. Removes the hook Job and its scheduling races without adding a new ordering dependency.

## Verification (`helm template`, repo's local kubevela chart + these registry values)

| Check | Default | With this change |
|---|---|---|
| `failurePolicy: Ignore` webhooks | 0 | **7** |
| `failurePolicy: Fail` residual | (Fail) | **0** |
| cert-manager `Certificate` / `Issuer` | 0 / 0 | **2 / 2** |
| `admission-webhooks/job-patch` cert-gen Jobs | present | **0 (removed)** |

Sub-chart hook Jobs already carry `helm.sh/hook-delete-policy: before-hook-creation,hook-succeeded` upstream, so no extra cleanup annotation was added.

## Rollout note (impact on live clusters)

All three clusters are currently healthy on the old config (`failurePolicy=Fail`, cert via Job). Syncing this switches the admission-webhook cert mechanism to cert-manager on **live** clusters — an admission-control change. Recommend syncing `kubevela-*` one cluster at a time (start with `spoke-dev`) and confirming `vela-core` stays 2/2 and the webhooks get a cert-manager-issued caBundle before rolling to hub/prod. Not auto-synced by this PR.


shapirov103 (approved):
LGTM

PR #922, opened 2026-09-22
Install/destroy resilience + automatic spoke OIDC

shapirov103:
Fixes accumulated from running the platform's install/destroy path end-to-end against successive failures, plus one feature: **arbitrary spoke names now get their Keycloak OIDC client and issuer wired automatically, with no per-spoke manual file.**

## Highlight: automatic per-spoke OIDC client generation
- Any spoke declared with a `clients:` entry in its existing `fleet/members/<name>/values.yaml` now gets its Keycloak client provisioned automatically — no separate `client-provisioner.yaml` file required, and no hardcoded spoke name anywhere in the chain.
- The externalSecret-mode fleet-secret (KRO spoke path) now emits `oidc_issuer_url` pointing at the hub's public IdP, so any spoke's JWT-validating workloads (e.g. agent-gateway) resolve a real issuer instead of an empty/invalid one.

## Install
- Gate install on CRD readiness instead of racing; stop silently masking required values; widen retry budget.
- Drop duplicate `resourcePrefix` keys in the KRO spoke instance and appset global block.
- Isolate cluster-provisioning RGDs so a disabled feature RGD (e.g. Ray) can never block spoke provisioning.
- Compute bcrypt without a Python dependency (portable `htpasswd`); report X-Ray status honestly instead of swallowing failures.
- Never export an empty `AWS_PROFILE`; fix a race in how broadcast env entries render under concurrent variable resolution.
- Remove pipefail landmines where a `grep` miss was silently aborting a step under go-task's strict shell.
- Make remote-git credentials optional when no external git host is configured (fixes a permanent Degraded state for GitHub-only installs), and fix the entry the gate was actually applied to.

## Destroy
- Never treat a CloudFormation tag alone as proof the workshop owns a VPC before deleting it.
- Select spoke VPCs by ownership tag in the sweep task's destroy calls.
- Fail loudly on a missing prerequisite instead of claiming unverified success.

## Tested

- **Live, on a running hub+spoke install (peeks-hub / us-west-2):**
  - `argocd-gitlab-repo-creds` app: confirmed the baked (post-sync) Application value, not just an isolated chart render — Synced/Healthy with no ExternalSecrets rendered when no git host is configured.
  - KRO spoke unblock: confirmed the isolated RGDs don't gate spoke provisioning — hub appsets (`kro-clusters-rgds-hub`, `kro-manifests-hub`, `kro-ray-rgd-hub`) all present and correctly scoped.
  - bcrypt fallback: `htpasswd -vb` confirms the generated hash verifies on a machine reproducing the original Python-mismatch failure.
  - pipefail landmines: reproduced the go-task abort (mvdan/sh strict pipefail) and confirmed the `awk`-based rewrite survives a grep miss.
- **By rendering (`helm template`) for the fleet-secret / OIDC change:**
  - `mode=externalSecret` with `hub.domain` set → `oidc_issuer_url` renders as `https://<domain>/keycloak/realms/<realm>`.
  - `mode=externalSecret` with `hub.domain` empty → annotation correctly omitted (no malformed `https:///...`).
  - `mode=direct` → unchanged, confirming no regression on the Crossplane/Terraform spoke path.
  - `client-provisioner.yaml` validated as well-formed YAML after the generator glob change (`yq`).
- **Not yet verified end-to-end:** the automatic-OIDC feature requires a companion change in the OAP overlay repo (consume `oidc_issuer_url` instead of a Keycloak-specific annotation name) before the full chain — client creation in Keycloak → issuer on the spoke secret → agent-gateway JWT policy → token acceptance — can be confirmed live.
- **Not yet run:** a full `task install` → `task destroy` → `task install` cycle from a clean account.


allamand:
## Validation on a fresh Workshop Studio event (kind-kro-ack path)

I deployed this branch end-to-end on a **fresh** Workshop Studio account (`deployment_attempts_count=1`, default VPC deleted) and verified the platform against each change family. Summary: **install succeeded in a single pass** (`InitScriptStatus=READY`), **3 EKS clusters ACTIVE** (hub + 2 spokes), **95/98 ArgoCD apps Healthy/Synced**. Details below.

> ⚠️ **Teardown / destroy resilience is NOT yet tested.** This run only exercised the *install* half. The whole `task destroy` path — boto3 preflight, the kept `delete-cluster` stderr, the new final "Destroy incomplete" check, the tag-based (`platform.gitops.io/prefix`) spoke sweep, and `HUB_CLUSTER_NAME` passed to `sweep-spoke-vpcs.py` — remains unexercised. That's the highest-value follow-up test.

### ✅ What validated

- **Automatic spoke OIDC + `insecure` threading** — both spoke cluster secrets carry `oidc_issuer_url` derived from the **hub's public domain** (`https://<hub-cloudfront>/keycloak/realms/platform`), not an in-cluster Service URL, and `insecure=true` is propagated hub→spokes. Exactly as intended.
- **RGD isolation (the structural fix)** — `kro-clusters-rgds-hub`, `kro-ray-rgd-hub`, and `kro-manifests-hub` are all Healthy/Synced; `eksclusterwithvpc.kro.run` is Active and all 11 RGDs are Active; both spokes provisioned. This confirms an optional/disabled capability can no longer stall the core cluster-provisioning wave chain.
- **GitHub-only resilience (gitlab-configured path)** — `argocd-gitlab-repo-creds` and `backstage` are Healthy/Synced. Note: the *interesting* branch of this fix (empty render / `optional: true` secret when **no** remote git is configured) is not exercised by the workshop path, which always configures GitLab.
- **Install resilience** — install completed in one pass with no manual retry, which is the headline goal of the Taskfile hardening.

### 🔴 Findings

**1. Kargo admin password hash is empty (install-resilience gap).**
`kargo-admin-credentials.ADMIN_ACCOUNT_PASSWORD_HASH` is **0 bytes**, both in the k8s Secret and in Secrets Manager (ESO reports `SecretSynced=True` — it faithfully synced an empty value). The seed produced an empty hash because **neither `htpasswd` nor an importable `python3` `bcrypt` was available on the IDE** when the seed step ran. The htpasswd-first change did its defensive job — the install no longer aborts (hence the clean single-pass) — but the graceful-degradation `else` branch left the hash empty, so **the Kargo admin UI login is broken** with only a warning in the log.
- *Recommended fix:* make the prerequisite present rather than only warning — add `apache2-utils` (`htpasswd`) to the IDE bootstrap image, or `python3 -m pip install bcrypt` at bootstrap. Optionally, in the workshop/SSM path treat an empty Kargo hash as a hard error (participants can't recover it themselves).

**2. Both `client-provisioner-<spoke>` apps are `OutOfSync` with SharedResourceWarning (side effect of this PR).**
The generator change from a per-member `client-provisioner.yaml` to the shared `fleet/members/*/values.yaml` means **both** spokes now generate a client-provisioner Application, and both claim the same hub-scoped Keycloak provisioner resources (`PodIdentityAssociation` / `Role` / `ServiceAccount` `peeks-hub-keycloak-client-provisioner`). Result: both apps are `Healthy`, op `Succeeded` ("all tasks run"), but permanently `OutOfSync` on a `SharedResourceWarning`. Cosmetic (realm-ensure still runs), but newly introduced by this PR.
- *Recommended fix:* de-duplicate ownership of the hub-side provisioner (single owning app), or mark the shared resources so ArgoCD tolerates multi-app ownership (e.g. `argocd.argoproj.io/sync-options: no-owner-ref` / drop them from the per-spoke render).

**3. `external-secrets-<spoke-prod>` is `Missing`/`Failed` (pre-existing, NOT caused by this PR).**
Sync fails with `CustomResourceDefinition clustersecretstores.external-secrets.io is invalid: metadata.annotations: Too long (>262144 bytes)`, then `namespace external-secrets not found` and `no matches for kind ClusterSecretStore ... ensure CRDs are installed first`. Spoke-**dev** is Healthy, and the app already has `ServerSideApply=true` — so this is a spoke-prod-specific **install-ordering race** (the CRD isn't Established before the ClusterSecretStore applies). This PR's `kubectl wait --for=condition=established crd/clustersecretstores.external-secrets.io` gate lives in the **kind/hub Taskfile** path, not in the ArgoCD-applied **spoke** ESO Application, so it doesn't cover this case. A re-sync should clear it.
- *Recommended fix (optional, separate from this PR):* extend the same CRD-established gate to the spoke ESO Application (sync-wave ordering or a PreSync wait), or rely on ArgoCD retry.

### Suggested follow-up tests (priority order)
1. **Teardown / destroy** — keep the event alive, trigger the CFN stack deletion with facilitator creds (do **not** terminate the event), and assert: no false "✓ Destroy complete" while the hub cluster survives; the new final check fails loudly if it does; boto3 preflight guards the sweep; zero residue (spoke VPCs by tag, `peeks/*` ECR repos, `/aws/eks/peeks-*` log groups, `peeks*` secrets).
2. Kargo admin login (finding 1) after adding htpasswd/bcrypt to the IDE.
3. De-dup decision for the shared client-provisioner resources (finding 2).
4. Re-sync `external-secrets-<spoke-prod>` and confirm it goes Healthy (finding 3).

_Environment identifiers (account ID, CloudFront domain) omitted; happy to share privately._

shapirov103:
Thanks for the fresh-account run. Both real findings held up, and one was my regression.

**Finding 2 (SharedResourceWarning) — fixed in `f75f0cd1`.** Your read was right. The ServiceAccount, IAM Role and PodIdentityAssociation are hub-scoped singletons. But the chart renders once per fleet member, so every member produced identical copies of all three. The Job and ConfigMap are named from `.Values.environment`, so they don't clash. That is why both apps looked Healthy yet stayed OutOfSync.

These three resources now live in the hub-only keycloak chart. It renders once, since `enable_keycloak` is set only for control-plane. It also already owns the `keycloak-config` Secret that the Job mounts as a volume, so the Job's pod cannot start until that Secret exists. Moving the ServiceAccount there adds no new ordering risk. One upgrade note: on existing clusters the old apps will prune these while the keycloak app creates them, so expect a brief window.

**Finding 1 (empty Kargo hash) — fixed in `895219dd`. This one was my regression.** Here is what happened. Main ran `pip install bcrypt`, then `python3 -c "import bcrypt..."`. I saw the real bug: bare `pip` often points at a different interpreter than `python3`. That mismatch was killing the last install step. But I fixed it the wrong way. I deleted the install and switched to `htpasswd`, instead of just changing the call to `python3 -m pip`.

On the Amazon Linux IDE, `pip` and `python3` are usually the same interpreter. So main installed bcrypt and got a real hash. My version only checked if bcrypt was already importable. It never tried to install it, so it wrote an empty hash. I replaced a working path with a broken one on the environment that matters most.

The gap in my testing: I only tested on a machine where the old code failed. I never tested one where the old code worked. Those are two different claims, and I had only proved the first.

The fix restores the install in the right form and removes the duplication. `scripts/bcrypt-hash.sh` is now the only implementation: htpasswd, then an importable bcrypt, then `python3 -m pip install bcrypt`. It exits non-zero instead of ever printing an empty hash. All four call sites now use it, so `errexit` handles the abort for free. Those four copies had already drifted apart in fallback order, messages and failure behavior. Net 30 lines removed from the provider Taskfiles. Verified by running it: the hash accepts the right password and rejects a wrong one, stdout holds only the hash, and the failure path aborts the task under go-task.

Your idea to add `apache2-utils` to the IDE image is still worth doing as a backstop. It lives in the CDK repo, not here. With the install restored, the hard-fail branch should almost never trigger.

**Finding 3 (spoke-prod ESO) — confirmed pre-existing, filed as #928.** You were right that this PR does not cover it. Every CRD gate I added runs in the provider Taskfiles against the hub, never against the spoke ESO app. I split it out because it is really two problems. The ordering race is cheap to fix with the PreSync `crd-wait-hook.yaml` pattern that `ack-pod-identity` already uses. But the real blocker is the CRD being rejected for `annotations: Too long`. That entry already sets `ServerSideApply=true`, so I cannot explain it from the config alone and it needs the failing cluster. Waiting on a CRD that never applies would not have helped anyway. The issue has a rough plan.

**Still untested:** teardown and destroy. I agree that is the top follow-up. Also unverified on a live cluster: the warning actually clearing, and Kargo accepting the hash at login.

allamand:
## Re-test of the two fixes on a fresh Workshop Studio environment

Re-tested `f75f0cd1` and `895219dd` on the same fresh account provisioned earlier (single `deployment_attempts_count=1`, `InitScriptStatus=READY`). The hub apps track the **branch** `fix/install-destroy-resilience` (not a pinned SHA), so ArgoCD auto-refreshed and pulled the new commits.

### Finding 2 — `f75f0cd1` (SharedResourceWarning): ✅ fixed

The SA/Role/`PodIdentityAssociation` are now owned solely by `keycloak-peeks-hub`:

- `ServiceAccount keycloak-client-provisioner` + Role + `PodIdentityAssociation keycloak-client-provisioner` → single owner, PIA `SYNCED=True / READY=True`.
- Both `client-provisioner-spoke-{dev,prod}` → **Healthy / Synced / Succeeded, no more `SharedResourceWarning`**.
- `keycloak-client-provisioner-prod` Job → **Complete 1/1**.
- **98/98 ArgoCD apps Healthy/Synced** (was 95/98 before the fix).

### Finding 1 — `895219dd` (empty Kargo admin hash): ✅ code correct + verified locally

- `scripts/bcrypt-hash.sh` reviewed: `htpasswd -B` first (with the `$2y$`→`$2a$` conversion for Kargo's Go bcrypt), then importable `bcrypt`, then `python3 -m pip install bcrypt` (the real fix) — and it **never emits an empty hash** (non-zero exit on empty input). All **4 call sites** (seed-secrets + kargo, in both kind-crossplane and kind-kro-ack) call it.
- Ran locally: valid password → a 60-char bcrypt hash that verifies (correct password accepted, wrong one rejected); empty input → exit 2 with no stdout.
- Runtime re-seed not exercised on this env: it already carries the old (empty) hash from the pre-fix provisioning, and re-running the seed goes through `task install` on the IDE, which I couldn't trigger in this session. A **fresh event built on the branch HEAD** would confirm "Kargo accepts the hash at login" end-to-end.

### Still open

- **Finding 3 / #928 (ESO spoke-prod)**: out of scope for these fixes; currently Healthy/Synced on this env (the ordering race resolved on a re-sync). The substance of #928 stands.
- **Teardown / destroy (family B)**: still **not tested** — this is the highest-value follow-up. The install path looks solid; the destroy-resilience half (boto3 preflight, `Destroy incomplete` final check, tag-based sweep) needs a real `delete-stack` on a live event to assert zero-residue.

allamand (approved):
LGTM

allamand:
### Runtime validation of `895219dd` (empty Kargo hash) — ✅ confirmed on a live env

Followed up on the one item left open from my previous comment (the Kargo admin hash was only verified at the code/local level). I re-seeded on the existing environment and it's now proven end-to-end:

1. Updated the IDE clone to branch HEAD (`0c8d12dd`) — the IDE-side scripts don't auto-refresh, so this pull was needed to pick up `scripts/bcrypt-hash.sh` + the updated call sites.
2. **"Before" state** matched the failure mode exactly: `<hub>/secrets` `.user_password_hash` was an **empty string**, and the IDE had **neither `htpasswd` nor an importable `bcrypt`** — i.e. the seed genuinely had to fall through to the last resort.
3. Re-ran `secrets-manager:seed-secrets`. Observed log:
   ```
   bcrypt-hash.sh: bcrypt not importable, installing it for this interpreter (python3 -m pip)
   ```
   → the `python3 -m pip install bcrypt` fallback fired as designed, and the hash went from empty → `$2b$10$…` (60 chars).
4. ESO synced it into `kargo-admin-credentials.ADMIN_ACCOUNT_PASSWORD_HASH` (60 chars), and I rolled `kargo-api` to pick it up.
5. **Login proof:** `bcrypt.checkpw(user_password, hash)` → **MATCH**. Admin login is functional; it was broken (empty hash) on this env because it was provisioned with the pre-fix code.

So the `895219dd` fix does what it promises: when no bcrypt provider is present, the pip fallback activates, a valid non-empty hash is produced, and it verifies against the workshop password. Combined with `f75f0cd1` (SharedResourceWarning gone, 98/98 apps Healthy/Synced), both fixes check out.

Only **teardown/destroy (family B)** remains untested — still the top follow-up.

allamand:
## Teardown / destroy test (family B) — results

Ran the destroy path end-to-end on a fresh account, keeping the Workshop Studio event **alive** (never `event terminate`) and triggering `delete-stack` directly with facilitator creds. The teardown went through the **CFN-delete → sweep CodeBuild** path (not SSM `task destroy`: the IDE resources are deleted early by CFN, so no SSM command runs). The sweep build clones `fix/install-destroy-resilience`, **SUCCEEDED**, and the stack reached **DELETE_COMPLETE**.

### ✅ What this PR validates (works)
- **Sweep-by-tag** — `Discovered owned spoke by tag: peeks-spoke-dev / peeks-spoke-prod` (the core of family B).
- **Resilient `delete-cluster`** — waits out `ResourceInUseException` (capabilities detached before delete) on hub + both spokes; all three deleted.
- **`git_clone_with_retry`** on the sweep clone.
- **Ordered teardown**: capabilities → CloudFront + VPC origin → ALB → RDS → 3 AMP scrapers → hub → spokes → 70 IAM roles → 8 secrets (force-deleted) → AMG workspace → 3 Logs delivery objects → 1 EIP.
- Provision in a single pass (`deployment_attempts_count=1`).

### ❌ Zero-residue NOT reached — 5 orphaned categories

| Residue | Root cause |
|---|---|
| **2 spoke VPCs** (+ 1 subnet each) + **2 EIPs** | **Timing**: `DependencyViolation` at sweep time (ENIs/NAT not yet released). Verified afterwards: NAT `deleted`, 0 ENI, 0 endpoint → a longer retry/wait would have removed them (same as the existing re-sweep loop already does for the IDE VPC). |
| **ECR `peeks-ray-vllm-custom`** | **`AccessDeniedException` on `ecr:DeleteRepository`** — permission missing from the shared role → `force-deleted 0`. |
| **S3 `peeks-ray-models-<acct>`** | **`AccessDenied` on `s3:DeleteBucket`** — same permission gap. |
| **4 log groups `/aws/eks/peeks-*`** | Sweep reports `No log groups to delete` → the sweep's log-group filter does not match these. |

### Note on the completeness guard
The **"Destroy incomplete"** guard was **not** exercised on this path — the sweep reported `Extended sweep complete` and the build was **SUCCESS** despite the 5 residues above. That check lives on the `task destroy` (Taskfile) side, not in the sweep script, so on the CFN-delete → sweep path teardown completeness is currently **not guarded** (a taggeD-resource residue does not fail the build).

### Actionable findings (tracked in follow-up issue)
1. **Spoke VPC delete needs a retry/wait** — the sweep gives up on the spoke VPCs at the first `DependencyViolation`; it needs the same ENI/NAT wait loop already used for the IDE VPC. High impact (2 VPCs + 2 EIPs orphaned every teardown otherwise).
2. **Shared role missing `ecr:DeleteRepository` + `s3:DeleteBucket`** — otherwise `peeks-*` ECR repo and the `ray-models` bucket are orphaned every time.
3. **Sweep log-group filter** does not match `/aws/eks/peeks-*`.
4. **`Destroy incomplete` not guarded on the sweep path** — the sweep should exit non-zero if tagged resources remain.

---

**Overall**: the install and destroy resilience improvements hold up well end-to-end — clean single-pass install, ordered/resilient teardown, sweep-by-tag working. The 5 residues are follow-up hardening items (timing retry + two IAM permissions + a filter + a completeness gate), none of which block this PR. From my side this **looks good to merge**; the residue items are captured as a follow-up for a separate change.

PR #925, opened 2026-09-23
fix(backstage): seed kro-clusters values.yaml for control-plane tenant (eks-cluster-template ENOENT)

allamand:
## Summary

Two related Backstage fixes, discovered while validating the rebuilt CNOE image:

1. **Point the default Backstage image to `:cnoe-1.1`** (`gitops/addons/configs/backstage/values.yaml`). The previous `:cnoe` image had a frontend regression where **clicking any scaffolder template rendered a client-side 404** ("PAGE NOT FOUND / someone dropped the mic"), while the template list page worked. The rebuilt `:cnoe-1.1` fixes it. Verified live on the hub (template pages load, scaffolder runs).

2. **Seed `kro-clusters/values.yaml` for the `control-plane` tenant** so the `eks-cluster-template` scaffolder no longer fails with ENOENT.

## Problem #2 — root cause (verified against the running image)

Running `eks-cluster-template` fails at the **`Update kro Clusters Configuration`** step:

```
error: The file .../tenants/control-plane/kro-clusters/values.yaml does not exist.
  at @roadiehq/scaffolder-backend-module-utils/dist/actions/merge/merge.cjs.js:154
```

The step uses **`roadiehq:utils:merge`**, which in **v4.1.2** (shipped in the CNOE image) deep-merges into an **existing** YAML file but **throws if it is missing** — it does not create it. The template's default tenant is **`control-plane`**, whose `spoke-values` dir had `crossplane-clusters/` but **no `kro-clusters/values.yaml`** (only `workshop` was seeded), so the default path never existed.

`roadiehq:utils:json:merge` *does* create the file, but it reads/writes **JSON** and would corrupt this YAML (and break the `workshop` tenant), so it is not a valid substitute.

## Changes

- `gitops/addons/configs/backstage/values.yaml`: `backstage_image` → `:cnoe-1.1`.
- Add `gitops/fleet/spoke-values/tenants/control-plane/kro-clusters/values.yaml` (`clusters: {}`), mirroring the `workshop` tenant, so `roadiehq:utils:merge` finds its target.
- Fix the misleading in-template comment that claimed `merge` creates the file.

## Notes

- The `eks-cluster-template` checks out the user's **fleet-config** repo (`fetch:plain`) and merges into it, so the same base file must also exist in fleet-config for existing installs (applied separately for the live hub). This PR ensures fresh installs ship it from the platform repo.
- Any tenant this template can target must ship a base `kro-clusters/values.yaml`; a follow-up could make the template tenant-agnostic (pre-create the skeleton conditionally). Seeding the default tenant is the minimal, safe fix here.


gitops/fleet/spoke-values/tenants/control-plane/kro-clusters/values.yaml
```diff
@@ -0,0 +1,22 @@
...
+# instead of Crossplane. Both can coexist.
+#
+# Example:
+# clusters:
+#   spoke-kro-test:
+#     tenant: control-plane
+#     environment: dev
+#     region: us-west-2
+#     k8sVersion: "1.35"
+#     accountId: "134136091485"
```
elamaran11: Remove Account IDs

elamaran11 (changes requested):
One Comment.

gitops/fleet/spoke-values/tenants/control-plane/kro-clusters/values.yaml
allamand: Good catch, thanks this is fixed

allamand:
Thanks , I've fix this

PR #23, opened 2024-09-10
Two New Backstage Templates for Developer Apps

elamaran11:
*Issue #, if available:*

*Description of changes:*
Two New Backstage Templates for Developer Apps which needs 
1. Gitea Repo and ArgoCD App
2. ArgoCD App fora GitOps Repo

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #24, opened 2024-09-14
cdk changes

renukakrishnan:
*Description of changes:*
CDK project for Codebuild 

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #926, opened 2026-09-23
feat(platform): add peeks-agent — PEEKS read-only agent instance

allamand:
## What
Adds `platform/peeks-agent/` — the **PEEKS-specific instance** of the OAP agent stack, deployable as a **single KubeVela Application** (one `kubectl apply`).

## Contents
- `deploy/peeks-agent-app.yaml` — one Application: agent + skills-mcp + eks-read-mcp + **chat-ui (incl. ConfigMap)** + **ACK `AccessEntry`** (hub + spokes, read-only ViewPolicy+AdminViewPolicy)
- `deploy/BUNDLE.md` — prerequisites, parameters, apply, MCP transport rules
- `skills-overlay/Dockerfile` — bakes `.kiro/skills` into the generic OAP skills-mcp image
- `buildspec.yaml` — CodeBuild: build + push the 4 pinned images

## Key design
- Reusable building blocks live upstream in **open-agentic-platform** (`a2a-chat-ui`, `skills-mcp`, `eks-mcp`, `strands-agent-base`); this dir only carries PEEKS specifics.
- **PEEKS skills reach the agent via the skills-mcp overlay** (`COPY .kiro/skills /payload/skills`), **not** the agent image — the agent is wired by config (`mcpServers`).
- eks-read-mcp reuses the agent SA to inherit its Pod Identity (no extra AWS association).

Depends on OAP images from awslabs/open-agentic-platform PRs #23/#24/#25.


shapirov103 (commented):
Solid work, and the single-Application packaging plus in-bundle access entries are
the right instincts. My comments are about the PEEKS/OAP seam rather than the
functionality.

## 1. Should the three images move here instead?

Agreed that the instance belongs in PEEKS: it is defined by PEEKS content and moves
at appmod's pace. The model I would like to settle on is that OAP is an external
platform publishing a spec for deploying agents, and PEEKS adheres to it.

If that is the model, the split in this PR runs the wrong way. `buildspec.yaml`
clones OAP and builds `applications/{a2a-chat-ui,eks-mcp,skills-mcp}`, which makes
PEEKS a consumer of OAP's source tree rather than its platform.

But looking at those three, none of them carry an OAP dependency:

- `eks-mcp` is supergateway plus `awslabs.eks-mcp-server`
- `skills-mcp` is a FastMCP server over a skills directory
- `a2a-chat-ui` is an A2A JSON-RPC client with Keycloak login

So why do they live in OAP at all? If they moved here, this becomes self-contained:
no cross-repo build, no waiting on OAP, and the skills server sits next to the PEEKS
skills it serves. OAP would keep only the spec we adhere to (the `agent` and
`mcp-server` ComponentDefinitions, and the published strands-agent image).

The counter-case is reuse: if these are meant as building blocks for every OAP
consumer rather than PEEKS scaffolding, OAP is the right home and OAP should publish
the images. Which of the two did you have in mind? Today OAP publishes only
strands-agent (`public.ecr.aws/oap/{skills-mcp,eks-mcp,a2a-chat-ui}` all 404), which
is why a per-environment build pipeline is needed at all.

## 2. The agent is already imageless, and that is right

Worth calling out as correct: `peeks-agent` sets no `image`, and the `agent` CD
defaults it to the published strands-agent
(`platform/oam/definitions/components/agent.cue:281`). The PEEKS agent is pure
configuration. Everything in point 1 is about the supporting services, not the agent.

## 3. Skills: image or configuration?

#24 says skills can arrive "via an overlay image or mount a volume/ConfigMap". The
ConfigMap path avoids coupling a `SKILL.md` edit to rebuild, repush, repin, reapply.
Was the overlay chosen for a reason I am missing? Note the bundle already uses the
ConfigMap approach for chat-ui, so the same problem currently has two answers.

## 4. chat-ui: the image is built, then overridden

The ConfigMap mounts over `/app/app.py` and `/app/static/index.html`, so the image
we build is not what runs. Two sources of truth, ConfigMap silently wins, and ~230
of this manifest's 406 lines become escaped Python and HTML with no linting or tests.

It also bypasses #23's "one image, any deployment" env-driven branding. Is there
something that path could not express? If the live UI simply drifted from the image,
landing the drift in the image seems cleaner.

## 5. These manifests look like cluster exports

`peeks-agent-app.yaml` carries live-state and defaulted fields:
`kubectl.kubernetes.io/restartedAt`, `schedulerName`, `terminationMessagePath`,
`dnsPolicy`, `revisionHistoryLimit`, `internalTrafficPolicy`, `ipFamilies`,
`defaultMode: 420`, and `finalizers: [group.ingress.k8s.aws/platform]` on the
Ingress. The committed finalizer is controller-owned and I would call that a bug.
Can this be hand-authored down to what is intentional?

## 6. Worth fixing

- **Image refs are not parameterized.** BUNDLE.md substitutes three placeholders, but
  the images are the literal `290085271972.dkr.ecr.../peeks-e2e/...`. Following
  BUNDLE.md points you at that account.
- **`:latest` on all three**, while BUNDLE.md prerequisite #4 says to pin real tags.
- **`eks-read-mcp` sets both `serviceAccount: eks-read-mcp` and
  `serviceAccountName: peeks-agent`** with different values. Which is intended?
- **`SESSION_SECRET` is a committed literal.** That is the session-cookie HMAC key,
  shared by anyone applying this file, and `app.py` already defaults to a random
  per-pod value. Secret, or drop it and take the default?
- **`KEYCLOAK_CLIENT_ID: mcp-client`** here, but `peeks-agent-chat` in `app.py` and
  BUNDLE.md. Also, is ROPC the right long-term choice for a browser UI?
- **The `eks-read-mcp` `command:` duplicates the image CMD verbatim**, including the
  read-only flags. Editing the Dockerfile would be silently overridden by the
  manifest. Drop it, or comment why it is deliberate.

## 7. Pending merges

No objection to these, just worth stating the order in the PR body: this depends on
#23/#24/#25 for the images, on #22 for the stock `agent` CD (so we can drop
`agent-fixed`) and for the agentgateway empty-host guard (so prerequisite #2 can
point somewhere other than a fork branch).

## 8. One manual step on the happy path

BUNDLE.md notes MCP tools load once at boot, so a fresh apply may need
`kubectl rollout restart rollout/peeks-agent`. That reads like platform behavior
rather than something an operator should have to know. Worth an issue against
strands-agent-base to retry or reconnect lazily, keeping the runbook note meanwhile.

## On eks-mcp

Not a complaint: the supergateway wrapper is justified, since
`awslabs.eks-mcp-server` is stdio-only (published docs show only
`uvx awslabs.eks-mcp-server` with `--allow-write` and
`--allow-sensitive-data-access`, no transport or port flag), and `0.2.1` is the
current latest.

One question: the launch config sets `FASTMCP_LOG_LEVEL`, hinting it is FastMCP-based,
and FastMCP speaks streamable HTTP natively (skills-mcp relies on that). Did you try
running it over HTTP directly? If it works, that removes supergateway, `--stateful`,
`sessionRouting: Stateful`, and the custom image together. Untested on my side, so
genuinely a question.

## Summary

Agreed on keeping the instance in PEEKS. The main thing I would like to settle is
point 1: whether the three generic services move here, or stay in OAP and get
published there. That plus the point 6 items would make this a pattern others can
copy.

allamand:
Thanks for this long and accurate review, I'll update according to your points

PR #927, opened 2026-09-23
fix(kro-clusters): stop requiring a real AWS account ID in git

shapirov103:
## What

The KRO spoke cluster chart required a real AWS account ID to be hardcoded into a fleet-config values file, with no fallback to a global value. This is inconsistent with the sibling Crossplane chart, which already resolves the account ID automatically from the hub's own identity.

## Root cause

`gitops/abstractions/kro/kro-clusters/templates/clusters.yaml`'s `accountId` and `managementAccountId` fields had no `| default $.Values.global.accountId` fallback, even though every other field in that file has one. Separately, `gitops/bootstrap/clusters-kro.yaml`'s ApplicationSet already resolves `.metadata.annotations.aws_account_id` for two other fields, but never exposed it as `global.accountId` for this chart to consume. The combination meant a spoke declaration had no way to inherit the account ID and had to hardcode it.

## Fix

Two files:
- `gitops/bootstrap/clusters-kro.yaml` — added `accountId: '{{ .metadata.annotations.aws_account_id }}'` to the values passed into the chart, using the same annotation-resolution pattern already used for the other two fields.
- `gitops/abstractions/kro/kro-clusters/templates/clusters.yaml` — changed:
  ```diff
  -  accountId: {{ $cluster.accountId | quote }}
  -  managementAccountId: {{ $cluster.managementAccountId | quote }}
  +  accountId: {{ $cluster.accountId | default $.Values.global.accountId | quote }}
  +  managementAccountId: {{ $cluster.managementAccountId | default $.Values.global.accountId | quote }}
  ```

This mirrors the sibling Crossplane path (`abstractions/resource-groups-kro/platform-cluster-kro/templates/claim.yaml`), which already defaults both `accountId` and `managementAccountId` to the same single `global.accountId`.

## Tested

`helm template` render output confirms both cases:

1. No per-cluster override (global fallback path) — `--set global.accountId=123456789012` with no per-cluster `accountId`/`managementAccountId` — rendered instance:
   ```
   accountId: "123456789012"
   managementAccountId: "123456789012"
   ```

2. Per-cluster override still wins (regression check) — same command plus `--set clusters.testcluster.accountId=999999999999 --set clusters.testcluster.managementAccountId=888888888888` — rendered instance:
   ```
   accountId: "999999999999"
   managementAccountId: "888888888888"
   ```
   confirms per-cluster values win over the global fallback (distinct from the global `123456789012`).

3. YAML validity: `yq eval '.' gitops/bootstrap/clusters-kro.yaml` exits 0 after the edit.

## Context

Found via review feedback on awslabs/open-agentic-platform PR #22, which had to commit a real AWS account ID into `gitops/fleet/spoke-values/tenants/default/kro-clusters/values.yaml` for exactly this reason, because there was no fallback available.


allamand (approved):

allamand:
LGTM

